Issue #3, opened 2015-04-03
Update R lesson so that it uses dataset on figshare


ethanwhite:
http://figshare.com/articles/Portal_Project_Teaching_Database/1314459

Moving this issue over from https://github.com/datacarpentry/datacarpentry/issues/128 following the closing of https://github.com/datacarpentry/datacarpentry/issues/117.

lwasser:
This is good - just a note that this lesson seems to use the  "messy version" of the data just like python does:
Surveys column names
record_id month day year plot species sex wgt

in the figshare version
record_id month day year plot species_id sex wgt

https://github.com/datacarpentry/planning/issues/3

we also need to add links to the data downloads up front.

PR #5, opened 2015-05-03
Update lessons so data come from figshare

fmichonneau:
This PR is to address #3, the code is updated to reflect the changes in the names of the columns with the new, cleaner dataset.

Once someone double checks, I'll also commit the corresponding HTML pages before merging.

Thanks!


fmichonneau:
I'm merging this so Kara can knit the lessons.

PR #8, opened 2015-05-06
updated sql lesson, edits, close con

lwasser:
hey @emhart -- i'm going to teach this tomorrow. i'm psyched so thank you again for putting it together. A few thoughts - some of which are implemented in draft in this PR to begin discussion.

**Nested Joins**
1. i think the join three tables lesson is important. do we want to include it here OR in the SQL lessons. @ethanwhite @fmichonneau  any thoughts on this? i think it's a very good thing to add to the SQL lesson. then we use it here as a "query and plot data" type of activity.
2. related to this, you don't really need aliases when you do this. All you have to do is call tableName.field. the aliases are nice sometimes but i think they add additional complexity that you don't need to learn a 3 way join and it's complex to introduce both the nested join AND aliases in one go.

So this works just fine:
SELECT species.genus, plots.plot_type, count(*)  FROM surveys 
JOIN species 
ON surveys.species_id = species.species_id
JOIN plots ON surveys.plot_id = plots.plot_id
GROUP BY plot_type, genus

I propose that we give them the basic syntax for a 3 way join and then make it a challenge. BUT ALSO maybe this belongs in the SQL specific lesson. Please weigh in.

**ALIASES**
I propose that we also cover aliases in the SQL lessons. It seems that pure sql lessons should be taught in SQL - then we should put things together in R - so in R, we have them create a query and then plot it or analyze results in some way. But we teach them about aliases, more complex  joins, etc in SQL.

My final thought is the scripted example is cool but it seems a bit odd in that - i'm not sure why i'd want to query every other year. This makes the lesson hard to follow. I wondered if we could think of another example. like maybe we want to create summary counts for every 5 or 10 years to look at longer term averages / trends?? or any other ideas?

Anyway - have a look, let me know what you think. I'll have more feedback after tomorrow post teaching!  
Cheers!

### DIFF
--- a/06-r-and-sql.Rmd
+++ b/06-r-and-sql.Rmd
@@ -27,7 +27,9 @@ Connecting R to sqlite databases
 
 R can connect to databases through a number of packages.  In our case we we will use [RSQLite](http://cran.r-project.org/web/packages/RSQLite/index.html) to connect to existing SQLite3 databases.  However you should be able to connect on almost any database in R via JDBC[http://cran.r-project.org/web/packages/RJDBC/index.html](http://cran.r-project.org/web/packages/RJDBC/index.html) or [ODBC](http://cran.r-project.org/web/packages/RODBC/index.html), or specific database packages (such as we are doing, or [MySQL](http://cran.r-project.org/web/packages/RMySQL/index.html) ).  
 
-To begin these exercises we'll connect to the mammal database you've just created.
+To begin these exercises we'll connect to the mammal database you've just created. We'll need the RSQLite library to be sure to install it first. `install.packages('RSQLite')`.
+
+It's also important that your working directory is set properly. Check it first `getwd()`. Set your working directory if it's not set to the same data directory in your data-carpentry folder that we've been working with.
 
 ```{r connect}
 library(RSQLite)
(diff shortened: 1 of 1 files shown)


lwasser:
one other note about aliases. I think they're best taught for column names. So have the students do count(*) and see the column name that's returned. Then do count(*) AS speciesCount or "species count" and get a nice clean table column name! it's a nice way to begin teaching the concept of aliases :)

emhart:
@lwasser I agree, I don't think we should introduce new SQL topics in this lesson.  I'm not sure how @ethanwhite feels about adding aliasing to the SQL lessons. In fact you need aliasing for one of the queries I wrote because it has a [sub-query](https://github.com/datacarpentry/R-ecology/pull/8/files#diff-ce39a380bb655f609e0964e34c329a0bL75) .  Aliases are also important because that's just how most people tend to write SQL so it will be important for students to understand it to decipher other people's SQL.

- Closing connections.  I'm not opposed to having it in there, however everything I've been able to find on the topic seems to point to the fact that it's good practice for making procedures readable, but doesn't impact the database at all.

- Yeah, every other year is a bit odd.  I wanted to come up with something that used a loop to generate queries though.  I think summary counts every 5 years is another good example though.

ethanwhite:
> I propose that we also cover aliases in the SQL lessons

I'm happy to add aliasing to the SQL lessons. Feel free to open an issue to remind me or send along a PR.

lwasser:
:+1:

ethanwhite:
Aliasing has been added to the SQL lessons in edd06f60c04fa3ea25b6fa8d375269bcb8c894f1.

hlapp:
@lwasser and @ethanwhite - I'm not sure what's next with this PR. Are we still considering this for merging, or are we factoring out more bits and pieces? Or should this be closed instead?

tracykteal:
Aliasing has now been added to the SQL lesson, and these are some nice updates to the R-SQL lesson, so I'll go ahead and merge.

Thanks @lwasser for the PR and @emhart for the comments and work on the lesson!

lwasser:
Wow - that is great!! i am glad this was resolved. sorry i really fell behind on helping with this repo this past few months.

hlapp:
@tracykteal should this be closed then? If not, it had conflicts that would need resolving before merging.

ethanwhite:
Hi @lwasser - terribly sorry that this hadn't gotten merged. We'd planned on doing it and then run into the merge conflict and it got lost in the pile. I just cleaned up the merge conflict locally and got it merged. Thanks again for the contribution.

PR #17, opened 2015-07-08
Applied lesson layout

apawlik:
Applied the lesson template to make it look like http://apawlik.github.io/R-ecology/

Some more cleaning up and lesson linking needed but at least the index page gets the logo and layout (roughly OK).


tracykteal:
The list of lessons still includes the Lesson 00 Name and Lesson 01 Name. I also think we want this list to be numeric.

Also, do we want to call this lesson 'Starting with R for data analysis for Ecology' or just 'R for Ecology'?

apawlik:
Sorry to mix in two things - the ggplot2 lesson adding and lesson styling. But at least the lessons are now listed correctly in index. I'm having problems generating html from Rmd for ggplot2 lesson. Don't know knitr :-(

apawlik:
OK. I managed to produce html for ggplot2 - PR landing soon

css/swc.css
```diff
@@ -73,7 +73,11 @@ body.stylesheet {
 div.banner {
     background-color: #FFFFFF;
     width: 100%;
+<<<<<<< HEAD
     height: 90px;
```
fmichonneau: There is a conflict here

index.md
```diff
@@ -11,43 +26,58 @@ import CSV files, the structure of data.frame, how to deal with factors, how to
...
+{% else %}
+{% include anySetup.html %}
+{% endif %}
 
-The lessons are written in Rmarkdown. A Makefile generates an html page for each
-topic using knitr. In the process, knitr creates an intermediate markdown
-file. These are removed by the Makefile to avoid clutter.
+<p><strong>Twitter</strong>: @datacarpentry</p>
 
-The Makefile also generates a "skeleton" file that is intended to be distributed
```
fmichonneau: If we remove this from here, we should make sure this info is included somewhere else.

apawlik: Maybe in README?

css/swc.css
apawlik: Ugh, sorry - though I checkout out the cleaned version. Let me fix this.

index.md
fmichonneau: Yes, that works. The Makefile would need to updated though, as currently README and index are kept in sync by it.

tracykteal:
I was just going to do the formatting and saw that this PR was already here and had been sitting for awhile. I'm just going to merge it as most of it looks good.

PR #24, opened 2015-08-14
a couple fixes to ggplot lesson to enable knitting and the HTML

tracykteal:
A couple things in the ggplot lesson weren't letting the document knit, so I fixed those and then cheated to render the HTML by just doing it in RStudio.


05-visualisation-ggplot2.Rmd
```diff
@@ -185,6 +185,7 @@ sections and points are counted wihin hexagons. The number of points per hexagon
 is encoded by color.
 
 ```{r}
+library(hexbin)
```
hlapp: If we need to load extra libraries for this call to work, that should go into into its own code chunk somewhere near the beginning. Unless you unload the library right after the plot invocation, but that would seem silly here.

05-visualisation-ggplot2.Rmd
```diff
@@ -152,7 +152,7 @@ ggplot(surveys_complete)
 - define aestetics (`aes`), that maps variables in the data to axes on the plot
      or to plotting size, shape color, etc.,
 
-```{r}
+```{r, eval=FALSE}
```
hlapp: So why does the command not work? Shouldn't we fix the command so that it will execute without error, instead of preventing it from getting executed but displaying the command in the rendered lesson anyway?

BTW you can also make R proceed despite error.

fmichonneau: Initially, I think the error was intended to show that using just the function `ggplot` wasn't enough to produce a plot and that a `geom_` function was needed.

05-visualisation-ggplot2.Rmd
fmichonneau: I agree. I think here we could use a different color palette to not have to call another package.

05-visualisation-ggplot2.Rmd
tracykteal: Right, it's in the code to demonstrate that it won't work. So, it's not supposed to work.

05-visualisation-ggplot2.Rmd
tracykteal: Ok, I'll modify to use another color palette and update the PR.

tracykteal: hexbin is required for that whole plot. It's not just a color scheme. Since it does require an additional library, I just took that whole example out.

tracykteal:
Comments on this PR have been addressed if someone wants to see if it's OK for merge.

tracykteal:
@karthik @fmichonneau could you take a look at this PR and see if it's OK to merge?

karthik:
Looks fine to merge although there are some conflicts to resolve.

This 

```r
species_counts <- surveys %>%
                  group_by(species_id) %>%
                  summarise(n=n())
```
could be replaced with

```r
species_counts <- surveys %>%
                  group_by(species_id) %>%
                 tally
```

to keep things simpler and emphasize built in functions.

tracykteal:
thanks, I'll go make those changes

05-visualisation-ggplot2.Rmd
hlapp: You can also set `error=TRUE` as a chunk option to have it executed, but not stop and instead display the error message, if you want to demonstrate that and how something fails.

fmichonneau:
I couldn't figure out how to merge this cleanly, so I recreated the changes in #37

PR #25, opened 2015-08-15
changed functions/arguments section to use round instead of read.csv in 00-getting-started

tracykteal:
In 00-getting-started we currently use 'read.csv' as an example of a function that has multiple arguments. This means though that we're loading in a file before we're ready to talk about loading in a file. As an instructor I've always stumbled over this section. I changed it to use 'round' instead, because it still talked multiple arguments but is easy to teach and understand.


00-before-we-start.Rmd
```diff
@@ -174,12 +187,8 @@ to do, and if you share your code with others that includes less known functions
...
 of the arguments for basic functions like `mean`, `min`, etc...)
 
-Another advantage of naming arguments, is that the order doesn't matter:
-
-```{r, eval=FALSE}
-read.csv(file="data/surveys.csv", header=TRUE) # is identical to:
-read.csv(header=TRUE, file="data/surveys.csv")
-```
+Another advantage of naming arguments, is that the order doesn't matter.
+This is useful when there start to be more arguments.
```
hlapp: Suggest to change _there start to be more_ to _there are more than a few_

hlapp:
Other than that minor change I suggested, looks great and ready to merge!

PR #32, opened 2015-09-18
Use the official combined file from figshare

ethanwhite:


ethanwhite:
I used `make` to update the html. This generated a really ugly diff, which I don't whether is expected or not having not used `knitr` much.

fmichonneau:
Thanks for the PRs Ethan! Yes, the ugly is normal. It's caused by knitr/Rmarkdown embedding the png files directly into the html that sometimes causes weird diffs.

PR #36, opened 2015-09-19
Use official combined dataset from repository

fmichonneau:
in `setup.R`.


tracykteal:
Thanks, definitely needed this update.

PR #39, opened 2015-10-12
Suggestions for ggplot lesson

hdashnow:
Hi,

[COMBINE](combine.org.au) taught the Data Carpentry R lessons in Sydney Australia last week (along with @lonsbio, @FPHeld, @steveped, @bobbledavidson and others). It was a big hit and attracted sponsorship from GigaScience, who are keen to get their own workshops up in China.

I had a bunch of suggestions for the ggplot2 lesson. Happy to split them up if you like some and not others. Briefly:

- Cut down the data cleaning/prep section. Students had only just learnt this, and still seemed bored hearing it again, even though we had just come back from a break. It also took valuable time from plotting and was already well covered in the previous session.
- Only create one cleaned data set `surveys_complete` and use it throughout the plotting exercises
- Notes on inheritance from the `ggplot()` call (although I didn't use the word inheritance)
- More consistency in the way plotting functions are used, particularly using `data = `, `x = ` etc. Explicit is better.
- A few minor fixes, spelling etc.

Let me know if you comments/thoughts on this.

Warm regards,
Harriet

### DIFF
--- a/05-visualisation-ggplot2.Rmd
+++ b/05-visualisation-ggplot2.Rmd
@@ -110,7 +110,7 @@ Make simple scatter plot of `hindfoot_length` (in millimeters) as a function of
 `weight` (in grams), using basic R plotting capabilities.
 
 ```{r}
-plot(x=surveys_complete$weight, y=surveys_complete$hindfoot_length)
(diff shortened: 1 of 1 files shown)


fmichonneau:
This looks great @hdashnow, Thanks!

I'm glad the workshop went well!

PR #42, opened 2015-11-01
Warn that need R >=3.1.2

willfurnass:
The workshop uses `dplyr`, which requires R>=3.1.2, yet some distribution repos (e.g. the default Ubuntu 14.04 repo) only provide older versions of R.  Added note in rSetup that might need to use CRAN packages to get R>=3.1.2.


hlapp:
Thanks for your contribution!

PR #41, opened 2015-10-13
Some further suggestions following the Sydney workshop

steveped:
Following the updates suggested by @hdashnow here are some further suggestions. Feel free to ignore or incorporate as you see fit.

(I've never forked or committed like this before, so let me know if I've done it wrong)

In brief:
- Vectors were more clearly introduced as containing all the same type of data.
- Subsetting was introduced for vectors first. This question spontaneously arose from the group at this point in the workflow
- Data frames are introduced as a collection of vectors
- Subsetting data frames was also introduced using the $ symbol, and treating each column as a variable

PR #44, opened 2015-11-22
Steveped's suggestions following the Sydney workshop

hlapp:
This is an attempt to remove the conflicting changes from @steveped's pull request #41. Here's the description quoted from there:

> Following the updates suggested by @hdashnow here are some further suggestions. Feel free to ignore or incorporate as you see fit.
> [...]
> In brief:
> - Vectors were more clearly introduced as containing all the same type of data.
> - Subsetting was introduced for vectors first. This question spontaneously arose from the group at this point in the workflow
> - Data frames are introduced as a collection of vectors
> - Subsetting data frames was also introduced using the $ symbol, and treating each column as a variable

Closes #41.


tracykteal:
Thanks @hlapp for this fix up and @steveped for the PR. This looks good and clarifies several points.

PR #45, opened 2015-11-23
misc fixes from going through the lessons

ctb:
I went through the lessons in preparation for teaching [the Caltech DC workshop](https://github.com/datacarpentry/2015-11-23-caltech/wiki) and fixed a few minor issues. My R and Markdown fu is not awesome so comments and suggestions would be welcome.

ping @tracykteal => comments/merge


tracykteal:
Thanks for the fixes, looks good. I'll go update the html.

PR #48, opened 2016-01-11
Fix layout

apawlik:
The old commits (listed 8 commits) should not affect the contents - see "Files changed"


fmichonneau:
OK, yes it looks like a minor change.

PR #49, opened 2016-01-11
Fixed link to the datasets

apawlik:


index.md
```diff
@@ -54,8 +54,10 @@ for each level and a very brief introduction to plotting.
 
 ## Data
 
-Data files for the lesson are available here: ({{page.dataurl %}})[{{page.dataurl %}}]
+Data files for the lesson are available here: 
+[{{page.dataurl %}}]({{page.dataurl %}})
 
+Please download the **.csv* (Comma Separated Value) files.
```
fmichonneau: I think there is a missing `*` after csv.

fmichonneau: Maybe add that 4 csv files need to be downloaded (species, plots, surveys, and combined).

apawlik: I thought about listing them explicitly and then got lazy :-) I'll do that.
About the missing * it's actually not it's supposed to be '*.csv' in italics in MD (so the forst star is formatting, second is wildcard and third is closing the formatting.

fmichonneau: Ha! I see about the stars :star2: But I think the wildcard needs to be escaped then... :grimacing: Otherwise use `*.csv` instead?

apawlik:
How about now?

fmichonneau:
Thanks, it looks good! :+1:

PR #51, opened 2016-01-29
Add challenge on vector types

adamobeng:
For instructor checkout process

### DIFF
--- a/01-intro-to-R.Rmd
+++ b/01-intro-to-R.Rmd
@@ -221,3 +221,7 @@ ones are lists (`list`), matrices (`matrix`), data frames (`data.frame`) and
 factors (`factor`).
 We are now going to use our "surveys" dataset to explore the `data.frame` data
 structure, which is one of the most common types of R objects we will see.
+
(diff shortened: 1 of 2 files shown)


fmichonneau:
Hi @adamobeng 

Thanks for your PR. I like the idea of the challenge but as it stands it's a little too advanced for data carpentry, as we don't cover integer and real data types (other than possibly mentioning them in passing). Also, it seems a little repetitive. It might be better to ask the question, "what happens if you mix types?" and have participants figure out the answer based on examples, and then have them reflect on what is happening. 

Finally, we use `purl` to extract the challenges into a script file that we distribute students (to reduce typing errors during workshops with many participants who have never programmed before). So, can you please update the PR so the code for the examples is self contain (and doesn't include the answer)?

adamobeng:
Hi @fmichonneau! Thanks for the feedback. I've taken your comments into account, and reformatted the challenge to match those elsewhere.

fmichonneau:
Thanks, it looks good!

PR #58, opened 2016-02-09
Link form index page to lesson 05 not working

kyrcha:
Fix the link: http://www.datacarpentry.org/R-ecology/05-visualisation-ggplot2.html that was giving a 404 error to http://www.datacarpentry.org/R-ecology/05-visualization-ggplot2.html


fmichonneau:
thanks for catching that!

PR #59, opened 2016-02-16
changes to 05-visualization-ggplot2.Rmd

esugis:
Added the modifications to the lesson 5 R-ecology.
Added cases:

- how to add white background to make graphs more readable
- how to change the orientation/size/colour of the axis text
- how to save plot to the file


tracykteal:
Thanks, both the white background and the text label orientation are things people commonly need to do, but not obvious to figure out. This is a nice addition.

PR #64, opened 2016-02-26
Added paragraph that defines commands and execution/running commands...

mikoontz:
… in the 00-before-we-start.Rmd lesson. Clarified in commenting section that ignoring everything on a line after a #  means that text is not executed. The intention with these edits is to create distinct processes out of writing commands and executing commands. In my opinion, this makes the notion of commenting more salient, and helps to explain how logistics are different when working in a script file versus working in the console.


fmichonneau:
Thanks for your contribution @mikoontz!

PR #65, opened 2016-02-29
Add challenge violin plot

markrobinsonuzh:
This is for my checkout session tomorrow ..
Cheers, Mark

### DIFF
--- a/05-visualization-ggplot2.Rmd
+++ b/05-visualization-ggplot2.Rmd
@@ -182,9 +182,14 @@ Notice how the boxplot layer is on top of the jitter layer? Play around with
 the order of geoms and adjust transparency to see how to build up your plot in
 layers.
 
-> ### Challenge
+> ### Challenges
 >
+> Replace the box plot with a violin plot (`geom_violin`)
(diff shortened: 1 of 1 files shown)


fmichonneau:
Hi @markrobinsonuzh 

Thanks for your contribution to the visualization lesson!

We cover the change in the orientation of the axis labels later in the lesson, and I think it's a little early to ask learners to do it at this point, so can you please remove the last question?

I like the idea of presenting violin plots and the log scale at this point of the lesson to emphasize how ggplots are assembled. Could you please add a sentence or two to motivate why we are asking for a violin plot and the log axis by mentioning both the technical side (how plots are assembled step-by-step, and the diversity of plots ggplot2 offers) and the data property side (large difference in the weight of the species).

Finally, please note that function names should include `()` (e.g., `geom_violin()`).

Let us know when you update the PR with these changes.

markrobinsonuzh:
Hi Francois,

Thanks for this.  I've made some modifications to the add-challenge-violin-plot branch.

Cheers, Mark

fmichonneau:
Thanks, it looks great!

PR #74, opened 2016-03-11
New challenge exercise for dplyr

stephlabou:
Rationale for exercise: keeping track of order of commands.

There is some flexibility in how this data frame is created (e.g. could select the columns, then filter by hindfoot_half value), but it's important to emphasize thinking through the flow when using multiple filter/select commands. (And later, if using multiple group_by()/ungroup()/group_by() in combination with other commands in a long chain.)

(Stephanie Labou PR for SWC/DC instructor training.)

### DIFF
--- a/04-dplyr.Rmd
+++ b/04-dplyr.Rmd
@@ -159,6 +159,16 @@ surveys %>%
 `is.na()` is a function that determines whether something is or is not an `NA`.
 The `!` symbol negates it, so we're asking for everything that is not an `NA`.
 
+> ### Challenge {.challenge}
+>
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks @stephlabou! It's a good addition as there is no challenge for this section of the lesson. Before I include it into the lesson, could you please add an answer? Look [here](https://github.com/datacarpentry/R-ecology/blame/gh-pages/04-dplyr.Rmd#L116-L123) for an example on how to format it.

stephlabou:
Answer included, submitting pull request. Glad it's useful!

PR #78, opened 2016-03-31
Added a challenge for subseting the values of a factor and creating a…

peterjuv:
… new factor.


fmichonneau:
Hi @peterjuv,

  Thank you for your contribution!

  The idea behind the challenge you are proposing is interesting but it falls out of scope for this lesson. We strive to make the data carpentry lessons of an appropriate length so they can be taught entirely during a workshop, and we are close to the limit of information we can have in these lessons. 

  Subsetting factors is not very different from subsetting vectors and we already spend quite a bit of time talking about it in the lesson. Therefore, if learners know how to subset vectors, they will be able to subset factors. One aspect that could be demonstrated that we don't currently have in the lesson is how to drop the unused levels of a factor after subsetting if you wanted to contribute that. If you do, please only submit the Rmd file you edit. Everything else is generated by the lesson maintainers at regular intervals. Don't hesitate to contact us if you have any questions!

PR #81, opened 2016-04-19
Update to 01-intro, add NA challenge

DrK-Lo:
I added an NA challenge in the "Missing Data" section that builds on the "Vectors and data types" section.  I also changed a few things in the order of the previous challenge, because I was confused about the way it was set up on a first read.

### DIFF
--- a/01-intro-to-R.Rmd
+++ b/01-intro-to-R.Rmd
@@ -207,19 +207,6 @@ factors (`factor`).
 
 ### Challenge
 
-* **Question**: We’ve seen that atomic vectors can be of type character,
-  numeric, integer, and logical. But what happens if we try to mix these types in
-  a single vector?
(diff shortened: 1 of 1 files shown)


fmichonneau:
Hi @DrK-Lo 

  Thanks for your contribution to the lesson, and bringing to our attention that the challenge on coercion needs to be re-arranged. We need to make it clearer the part that goes with the lesson (i.e., for the instructor) and the part that goes in the handout for the participants.

  I like the idea of having a challenge for the missing data section. The challenge you propose is slightly advanced in the sense that the participants will need to draw both on the concept of coercion, and missing data to understand what is going on. Could you please add another small challenge before the one you propose? Something that would only draw on the concept of NA and possibly that would demonstrate how the function `is.na()` or `na.omit()` works?

  Thank you, and don't hesitate to let us know if you have any questions.

PR #82, opened 2016-04-26
Name of Submitter: Akima George

ASGeorge:
This is the exercise submission for the SWC/DC instructor training checkout- item 1.


fmichonneau:
Hi @ASGeorge 

  Thanks for your contribution. During our workshops we try to limit information overload by avoiding the repetition of information. Additionally, this lesson is already nearing the maximum amount of information it can contain. We demonstrate how to do joins in the SQL lessons where it fits better. If workshop participants understand this concept, they will be able to use it with dplyr later on.

I'm closing your PR, but don't worry it fulfills the requirement on your path to become an instructor.

Don't hesitate to open issues or pull requests in other parts of the lesson if you'd like.

PR #86, opened 2016-05-25
Update learning objectives and text in intro(00), and revise content in ggplot (05).

ashiklom:
# Lesson 00: Before we start
* Original learning objectives were very vague. I tried to make them more specific by tying them to precise tasks learners should be able to accomplish.
* The organization of some of the text did not feel very logical. I revised it to try to introduce ideas in a more natural way, and to elaborate somewhat on a few key ideas.

# Lesson 05: Visualization with ggplot2
* A key strength of ggplot is the ability to incrementally build on existing plot templates, which are stored as ggplot objects. I revised much of the tutorial to showcase this ability, which I believe makes the tutorial easier to follow *and* makes ggplot more useful to learners in the long run.
* I also added a few minor details about ggplot capabilities related to the current topics (for instance, `facet_grid` in addition to `facet_wrap`, and the ability to set aesthetic mappings outside the `ggplot()` call).

NOTE: These suggested revision are related to my completion of the SWC Instructor Training.


fmichonneau:
Thanks for suggestions @ashiklom!

I like the changes you made to the 00 lesson.

For the ggplot2 lesson, we don't use variables to build the plots in the lesson, because when we are teaching it, it allows to modify progressively the original call in a way that mimic how ggplot2 are typically assembled. I like what you added for the themes, and it looks like a good place to demonstrate the concept of using variables so I'll keep it here. I'm going to merge your changes and edit them directly.

fmichonneau:
Also, for your future contributions, please only push changes to the Rmd files. The html files are generated by the lesson maintainers. Thanks!

PR #90, opened 2016-05-31
Minor revisions to data.frame lesson

ashiklom:
1. I removed the reference to the `cbind` and `rbind` commands because they
   produce counterintuitive results that are probably not worth discussing
   here. Specifically, both functions produce matrices, meaning that they
   coerce vectors to a common data type. In the given example, this means that
   both `colors` and `counts` columns are converted to factors, which is
   **not** intuitive behavior. Therefore, I personally strongly advise against
   using `cbind()` to make `data.frame`s from scratch (though there's nothing
   wrong with using these functions to add rows or columns to an existing
   `data.frame`, since that uses a different method of `cbind`).

2. I added some comments and (brief) mention of the fact that different syntax
   for subsetting a `data.frame` produces different results. Personally, I
   think it would be better to introduce the concept of a `list` in R first and
   then show that a `data.frame` is a special case of a `list` where all of the
   elements have the same length -- that would clarify some of the
   counterintuitive aspects of `data.frame` syntax -- but that would require
   substantial reworking of this lesson, so I'll leave it as just a thought.

3. The double underscore notation for boldface wasn't rendering correctly in
   the HTML, so I replaced all instances of it in this modoule with double
   asterisks. This should probably be done in the other lessons at some point
   (or the build script revised to accommodate both), but I haven't gotten
   around to it yet.

### DIFF
--- a/03-data-frames.Rmd
+++ b/03-data-frames.Rmd
@@ -87,12 +87,12 @@ fix them? Don't hesitate to experiment!
                                   year=c(1942, 1970))
     ```
 
-1. Can you predict the class for each of the columns in the following example?
+2. Can you predict the class for each of the columns in the following example?
    Check your guesses using `str(country_climate)`:
      * Are they what you expected?  Why? Why not?
      * What would have been different if we had added `stringsAsFactors = FALSE` to this call?
      * What would you need to change to ensure that each column had the accurate data type?
-
+ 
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks for your contribution @ashiklom. It looks good.

The workshop is intended for beginners and we already introduce a lot of concepts so we don't have room for lists as it stands.

PR #91, opened 2016-06-02
Byrnes edit of data frames to include filtering.

jebyrnes:
In this branch I

1) Added some additional material on `-` notation in subsetting data frames, as I find this incredibly common - I use it all the time.
2) Fixed one of the subsetting exercises.
3) Added two new subsetting exercises, one of which uses the `-` notation, and both of which build on one another, in addition to looping in `nrow()` concepts from earlier lessons.


fmichonneau:
Thanks @jebyrnes !

PR #100, opened 2016-07-10
Elaboration of '=' vs '<-'

tylerjrichards:
Assignment operator
<- is the assignment operator. It assigns values on the right to objects on the left. So, after executing x <- 3, the value of x is 3. The arrow can be read as 3 goes into x. For historical reasons, you can also use = for assignments, but not in every context. Because of the slight differences in syntax, it is good practice to use always <- for assignments, except when specifying the values of arguments in functions, when only = should be used, see below.
+The Google R Style Guide describes the identical result, but does so because your code may still work if you use '=' but will confuse others who attempt to understand and replicate.


fmichonneau:
Hi @tylerjrichards 

Your pull request is coming with a lot of commits that should already be incorporated into the gh-pages branch, and it makes it very difficult to review your contribution as there are a couple of months of diff with it.

Could you please pull all the recent changes from this repository into yours, and re-commit just your changes as a new pull request (only the Rmd file) so we can review it?

Don't hesitate to contact me if you need assistance.

Thanks for your contribution!

tylerjrichards:
hey @fmichonneau 
I am not sure how exactly to do that. I see now where I went wrong, but am not sure how to do it correctly. If I see any issues with the lesson, how do I commit changes to the correct place?

let me know,

thanks

fmichonneau:
I'm not sure what happened in your case but I suspect that some modifications from the data carpentry fork weren't included in your repository.

Assuming that you've already forked the repository and cloned its content locally, and that you have 2 remotes set up (`origin` for your fork, and `upstream` for the Data Carpentry one, you can check by running `git remote -v`), the general workflow should be:

1. make sure you are on the gh-pages branch and that origin and remote are even:

   ```
   git checkout gh-pages 
  git pull upstream gh-pages
  git push origin gh-pages
  ```

2. create a branch for your fix

   ```
   git checkout -b my-fix
   ```

3. Do the modifications to the Rmd files and commit them

    ```
   git add *.Rmd
   git commit -m "fix something that was broken"
   ```

4. Push to your fork

   ```
   git push origin my-fix
   ```

5. Go on Github and visit either your fork  or data carpentry's and open the pull request. Make sure only your commits show up when doing so.

For this pull request, if you still have the branch in which you did your modifications, then you might be able to do:

```
git checkout instructor_notes
git pull upstream gh-pages
# here git might ask you to write a commit message to reflect the merges
git push origin instructor_notes
```

Let me know if it helps, or if you encounter any errors.

Cheers!

PR #107, opened 2016-08-02
Fix some typos

iimog:
Fixed a couple of typos. The `wgt` to `weight` change is important.
As is the addition of the missing backtick which broke styling.
I kept each fix in a single commit so you can cherry pick but I can also stash them if desired.


fmichonneau:
thanks for catching these typos!

PR #111, opened 2016-08-12
Simplify and clarify factor-ordering challenge

ashander:
Introducing `table` does not seem necessary, as the main learning goal is to understand and manipulate factors.

Also: This is for SWC instructor training checkout from 2016-02-22-training-ucdavis


fmichonneau:
Thanks @ashander , that looks good!

PR #115, opened 2016-09-17
adding-challenge-05-visualization-ggplot2.Rmd

clarashaw:
Added a bullet point (L203-206) to the challenge about boxplots. This extended challenges requires students to use previous knowledge about integers and factors to assign colors to the graph correctly.


fmichonneau:
I think that's a useful addition, thanks!

PR #117, opened 2016-09-30
Capitalisation of RStudio and Rstudio

lmsimp:
When reading the section in the first lesson ["Before we start"](https://github.com/lmsimp/R-ecology-lesson/blob/help-info/00-before-we-start.Rmd) I have found inconsistent use of the capitalisation of RStudio. I have fixed this in this PR. 

I have also added how to customise the pane layout of RStudio. I found that some attendees don't always have the default setup and knowing how to change this to match the instructors setup (if they so wish) is useful. As this is only adding a few words after the first mention of customisation of the panes, I included it in this branch as it is part of the same section as the typos I found. Hopefully this is clear in the two commits in this branch of this PR.

This is also to fulfil part I of the checkout procedure for the DC instructors training (attended Cambridge 19/20th September).


fmichonneau:
:+1:

PR #118, opened 2016-09-30
Added information write reproducible examples

lmsimp:
In the first lesson towards the end we talk about how to ask for help from the community e.g. Stackoverflow, R mailing list etc. I have found a few small typos and also added two links on [How to write a reproducible example](http://adv-r.had.co.nz/Reproducibility.html) (this is a nice article from Hadley Wickham) and [How to ask a question on Stackoverflow](http://stackoverflow.com/help/how-to-ask) as I thought this would be a useful reference for newbies.

This is also to fulfil part I of the checkout procedure for the DC instructors training (attended Cambridge 19/20th September).


fmichonneau:
Thank you for these edits! They are helpful.

PR #119, opened 2016-10-03
Modifying learning objectives lessons 04, 05, 06

lachlandeer:
In this PR I modify the learning objectives for lessons

04-dplyr
05-ggplot2
06-SQL

The main objective was to add some clarity by elaborating on each objective so that a non expert could read them and hopefully understand what they are going to do without needing to know only R lingo. In essence these are slight extensions on the existing objectives 

Best

Lachlan

PS. Pull request for Instructor Traning, Data Carpentry


fmichonneau:
it looks good to me, thanks!

PR #124, opened 2016-10-18
change "error" to "warning"

### LINKED ISSUE #123: referring to a warning as an error, should we change?
At the very bottom of 01-intro-to-R.html, there's a challenge where learners are asked to take the mean of a string vector that "appears" to contain numbers. As desired (by us) the command returns a warning, but the text refers to it as an "error". There is mention earlier in the lesson that learners should "Try to use the correct words to describe your problem" (and of course there are real differences) so I think it would be consistent to change the text to "warning". If people agree, I'll submit a PR.

mfoos:
Fixes #123

### DIFF
--- a/01-intro-to-R.Rmd
+++ b/01-intro-to-R.Rmd
@@ -364,14 +364,14 @@ heights[complete.cases(heights)]
 
 ### Challenge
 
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks!

PR #131, opened 2016-11-16
Creation of the "data" directory now explicit -- was part of a silent…

ksamuk:
… source(). If using this lesson in isolation, allows the file download in the next section to be successfully executed.


fmichonneau:
Thanks for this @ksamuk. As mentioned in #131, this handled by the setup.R script and I prefer to have a single script used across all lessons. This lesson is intended to be taught in its entirety and therefore the folder should have been created in the first lesson ("Before we start").

PR #133, opened 2016-11-24
missing values read in

kdirek:
This is being submitted as part of instructor training.


fmichonneau:
Thanks, that's a useful addition.

PR #138, opened 2016-12-01
Removed the first sentence after loading pacakges,

jonkeane:
It looks like this sentence referenced a plot that was removed a few commits ago.


fmichonneau:
thanks for catching this!

PR #140, opened 2016-12-02
Edited learning objectives for episode 4

cvoter:
I understand that a goal across SWC and DC lessons is to make learning objectives more measurable and framed in terms of observables. I tried to do that here using verbs that fit more clearly into Blooms taxonomy, rather than 'know' or 'be able to'.


fmichonneau:
that looks good to me. Thanks!

Issue #155, opened 2016-12-21
Lesson 01-intro-to_R.Rmd consistency 


ytakemon:
This is a minor inconsistency that I noticed in the first challenge regarding data types. The challenge R code uses single quotes whereas the lesson uses double quotes throughout. Since R will also display double quote when encountering characters, the challenge code should be kept consistent.

```
num_char <- c(1, 2, 3, 'a')

num_logical <- c(1, 2, 3, TRUE)

char_logical <- c('a', 'b', 'c', TRUE)

tricky <- c(1, 2, 3, '4') 
```
Should be:

```
num_char <- c(1, 2, 3, "a")

num_logical <- c(1, 2, 3, TRUE)

char_logical <- c("a", "b", "c", TRUE)

tricky <- c(1, 2, 3, "4") 
```

PR #158, opened 2016-12-29
Fix consistency with use of quotes [#155]

kemya:


ctb:
This PR looks good to me! (@kemya, note that I'm not a lesson maintainer so I'm chiming in to say it's good, even though I can't/wouldn't merge myself!)

fmichonneau:
Thanks for spotting and fixing this @kemya !

PR #160, opened 2017-01-01
R ecology hexbin

fboehm:
Hi, I added a challenge question about using hexagonal binning in the visualization component of R-ecology lesson. I hope that it is useful & usable. If I should revise the exercise or reformat it, please let me know.

Thanks,
fred

### DIFF
--- a/05-visualization-ggplot2.Rmd
+++ b/05-visualization-ggplot2.Rmd
@@ -116,7 +116,7 @@ surveys_plot
 
 > ### Challenge
 >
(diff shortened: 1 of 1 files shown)


fmichonneau:
Hi @fboehm 

Thanks for your contribution! We used to have a demonstration of `geom_hex` as part of the lesson, but we dropped it, in part, because of the requirement of having to install an extra package.

Could you please make the challenge optional (just add `(optional)` after Challenge?

thanks!

fmichonneau:
thanks!

Issue #126, opened 2016-10-21
rewrite SQLite lesson to use dplyr to interact with database


fmichonneau:
Easier syntax and more coherent with the rest of the lesson. RSQLite package still needs to be installed.

Issue #121, opened 2016-10-04
No mention of downloading "portal_mammals.sqlite" file in lesson 06 R and SQL


cwmohr:
It is my impression that one should be able to teach more or less directly the R-ecology-lessons. This is not a problem for lesson 00 - 05. However, it is a bit more difficult for lesson 06 "R and SQL", because the requirement of specific files:

In lesson 06 [(SQL databases and R)](http://www.datacarpentry.org/R-ecology-lesson/06-r-and-sql.html) , under the section _"Connecting R to sqlite database"_ there is the following R-script:

``` r
## Set dbname and driver out of convenience
myDB <- "data/portal_mammals.sqlite"
conn <- dbConnect(drv = SQLite(), dbname= myDB)
```

There is however no earlier mention for downloading the file `"portal_mammals.sqlite"`.

I therefore suggest that lesson 06 be updated, by taking into use the `download.file()` function:

``` r
## download mammals database to data folder.
download.file("https://ndownloader.figshare.com/files/2292171", "data/portal_mammals.sqlite")
```

PS: All the file can be downloaded from http://dx.doi.org/10.6084/m9.figshare.1314459.

fmichonneau:
discussion for this issue continues in #162

PR #162, opened 2017-01-09
First proposal for a dplyr-database lesson, addressing issue 126

tomsing1:
This is a first proposal with 07-dplyr-and-databases.Rmd, an alternative to / replacement of lesson `06-r-and-sql.Rmd` It addresses issue / contribution request #126 . I am happy to improve the proposal, if it is considered useful at all. Any feedback is very welcome.

### Some quick observations:

- I tried to stick to the existing lesson 06 as closely as possible, but using dplyr shifts the learning objectives from writing SQL to understanding the concept of lazy evaluation in R.
- The most difficult challenges of lesson 06 turned out to be very straightforward using dplyr's notation.
- There were some reference to a GUI (firefox plugin) to examine SQLite databases, which I believe are largely outdated by now. I have removed them, but am happy to add them back in if necessary. 
- The database creation section also leverages dplyr's `copy_to` command. Like the original version that directly used RSQLite, the example is not suitable for very large datasets, as it is funneled through R.
- Finally, in my opinion connecting to databases via dplyr feels like a much more natural next step 
given the previous lessons in this course. In this proposal, there is no need for learners to understand SQL, although an example of the translation that happens behind the scenes is provided.

### DIFF
--- a/07-dplyr-and-databases.Rmd
+++ b/07-dplyr-and-databases.Rmd
@@ -1,0 +1,340 @@
+---
+title: "SQL databases and R"
+author: Data Carpentry contributors
+---
+
+```{r setup, include=FALSE}
+library(knitr)
+opts_chunk$set(
+  message = FALSE, 
+  echo = TRUE,
+  purl = FALSE
+)
+```
+
+```{r source, include=FALSE}
+source("setup.R")
+if (file.exists("portalR.sqlite")) file.remove("portalR.sqlite")
+```
+
+> ## Learning Objectives
+>
+>  By the end of this lesson the learner will:
+>
+> * Access a database from R
+> * Run a SQL query in R using dplyr.
+> * Create an SQLite database from existing .csv files.
+
+------------
+
+# Introduction
+
+So far, we have dealt with small datasets that easily fit into your computer's
+memory. But what about datasets that are too large for your computer to
+handle as a whole? In this case, storing the data outside of R and organizing it
+in a database is helpful: Connecting to the database allows you to retrieve only
+the chunks needed for the current analysis.
+
+Even better, many large datasets are already available in public or private
+databases. You can query them without having to download the data first.
+
+R can connect to almost any existing database type. For example, the
+[dplyr](https://cran.r-project.org/web/packages/dplyr/index.html) package you
+used in the previous lesson supports connecting to the popular open source
+databases [sqlite](https://sqlite.org/), [mysql](https://www.mysql.com/) and
+[postgresql](https://www.postgresql.org/), as well as google’s
+[bigquery](https://cloud.google.com/bigquery/).
+
+```{r dplyr}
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks for doing that @tomsing1! It looks great :smile: 

For this lesson, I would still like to show how to use the SQL syntax with dplyr to tie it back to the SQL lesson. It's useful to demonstrate that the dplyr syntax can directly be used on the database, but it seems also useful to show that you can send directly SQL queries to the database. So, it would be great if you could add a paragraph to show how to do that. It might also be interesting to contrast how much simpler it is to do the challenge using the pure dplyr syntax rather than mixing R and SQL to build the query.

We still use the firefox plugin to teach the SQL lesson, so it would be useful to keep references to it in the lesson.

Thanks again!

tomsing1:
Thanks a lot for the feedback. I will implement the changes you suggested - great ideas!

fmichonneau:
I'm cleaning up issues, and there is one (#121) that indicates that we should include a link to download the SQL database from figshare.

Typically, this lesson is taught right after teaching SQL, so the learners have the file readily available (it's generated during the SQL lesson), but I can imagine that in some cases, the SQL lesson is skipped all together and the database isn't generated by the participants.

I'm going to add a note in your pull request, and it would be great if you could incorporate it. Thanks!

07-dplyr-and-databases.Rmd
```diff
@@ -0,0 +1,340 @@
...
+library(dplyr)
+```
+
+# Connecting to databases from R
+
+We will continue to explore the `survey` data you are
+already familiar with from previous lessons. Instead of reading the data from a 
+csv file, we will connect to an SQLite database with the same information
+(and more). The SQLite database is contained in a single file:
+`portal_mammals.sqlite` and we point R to it with dplyr's `src_sqlite()`
```
fmichonneau: The SQLite database is contained in a single file `portal_mammals.sqlite` that that generated during the SQL lesson. If you don't have it, you can download it from Figshare using:

```{r download-database, eval=FALSE}
## download mammals database to data folder.
download.file("https://ndownloader.figshare.com/files/2292171", "data/portal_mammals.sqlite")
```
We can point R to this database with dplyr's `src_sqlite()` command.

tomsing1:
Thanks a lot for your input and feedback @fmichonneau ! I have updated the lesson and incorporated your suggestions - except for referencing the GUI. I wasn't quite sure where that would fit in, especially as the lesson is pretty long already. 

In the current R-ecology curriculum, learners progress from using dplyr on local datasets to querying databases. So it seems they are already used to scripted analyses. Perhaps discussing the trade-offs of GUIs vs scripted queries could be done in the SQL lesson? Any feedback is very welcome!

fmichonneau:
That looks good! Thanks for doing this!

PR #172, opened 2017-01-13
Revise lesson about factors

fmichonneau:
This is a first iteration of cleaning up the section about factors. I moved up the section about subsetting data frames as we need it to work with the sex column from the surveys data.


tracykteal:
Thanks @fmichonneau. It does make more sense to move these to the date frames section. I find that's where it's more natural to introduce these concepts.

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,132 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+column `"day"` by typing `surveys$d`. It's a shortcut, as with all shortcuts,
+they can have dangerous consequences, and are best avoided. Besides, with
+auto-completion in RStudio, you rarely have to type more than a few characters
+to get the full and correct column name.
+
+> ### Challenge
+>
+> 1. Create a `data.frame` containing only the observations from row 1999 of the
+>    `surveys` dataset.
+>
```
tracykteal: Because we have year data in this data frame, using 1999 is a little confusing, and sometimes people are solving it wrong. How about changing this to '100' or '3000' or something less year-like?

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,132 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+> 2. Notice how `nrow()` gave you the number of rows in a `data.frame`? Use
+>   `nrow()` instead of a row number to make a `data.frame` with observations from
+>   only the last row of the `surveys` dataset.
+>
```
tracykteal: I wonder if we can build to the solution a little more here. e.g.

Notice how 'nrow()' gave you the number of rows in a 'data.frame'. 
- Use that row number to pull out just the last row of the data.frame. 
- Compare that with what you see as the last row using 'tail' to make sure it's meeting expectations. 
- Now pull out that last row using 'nrow()' instead of a row number.
- Instead of having it output to the screen, create a new data.frame from that last row

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,132 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+> 3. Use `nrow()` in conjunction with the `:` function to create a new
+>    `data.frame` called `surveys_last_100` that includes the 100 rows of the
+>    survey data frame.
+>
+> 4. Combine `nrow()` with the `-` notation above to reproduce the behavior of
+>   `head(surveys)` excluding the 7th through final row of the `surveys` dataset.
+
```
tracykteal: Suggest changing to

Combine `nrow()` with the `-` notation above to reproduce the behavior of `head(surveys)` keeping just the first through 7th rows of the `surveys` dataset.

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,132 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+###   last row of the `surveys` dataset.
+###
+### 3. Use `nrow()` in conjunction with the `:` function to create a new
+###    `data.frame` called `surveys_last_100` that includes the 100 rows of the
+###    survey data frame.
+###
+### 4. Combine `nrow()` with the `-` notation above to reproduce the behavior of
+###   `head(surveys)` excluding the 7th through final row of the `surveys` dataset.
+###
+```
```
tracykteal: This is a repeat of the above challenges so that it ends up in the Rmd code document that students have. So, this would just have to be updated to reflect any challenge changes.

02-starting-with-data.Rmd
```diff
@@ -123,6 +245,11 @@ useful but not necessarily intuitive, and therefore require some attention.
 ### Factors
 ```
 
+When we did `str(surveys)` we see that several of the columns consist of
+integers, however, the columns `genus`, `species`, `sex`, `plot_type`, ... are
+of a special class called a `factor`. Factors are very useful but not
+necessarily intuitive, and therefore require some attention.
+
```
tracykteal: suggest

Factor are very useful and are actually something that make R particularly well suited to working with data, so we're going to spend a little time introducing them.

02-starting-with-data.Rmd
```diff
@@ -197,40 +323,55 @@ Notice that in the `levels()` approach, three important steps occur:
...
+```{r, purl=TRUE}
+## bar plot of the number of females and males captured during the experiment:
+plot(surveys$sex)
+```
 
+In addition of males and females, there are about 1700 individuals for which the
+sex information hasn't been recorded. Additionally, for the these individuals,
+there is no label to indicate that the information is missing. Let's rename this
+label to something more meaningful. Before doing that, we are going to create a
+new object so we don't loose the original data:
```
tracykteal: suggest something like, instead of Before doing...

'We're going to pull out the data on sex and work with that data, so we're not modifying the working copy of the data frame'

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,132 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+    * `rownames()` - returns the row names
+
+* Summary:
+    * `str()` - structure of the object and information about the class, length and
+	   content of  each column
+    * `summary()` - summary statistics for each column
+
+Note: most of these functions are "generic", they can be used on other types of
+objects besides `data.frame`.
+
```
tracykteal: Might want to add the code here for actually doing these things, e.g. str(surveys), so that instructors know they should be typing these with learners rather than just talking about them.

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,143 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+surveys$species_id          # Result is a vector
+```
+
+For our purposes, the last three notations are equivalent. However, the last one
+with the `$` does partial matching on the name. So you could also select the
+column `"day"` by typing `surveys$d`. It's a shortcut, as with all shortcuts,
+they can have dangerous consequences, and are best avoided. Besides, with
+auto-completion in RStudio, you rarely have to type more than a few characters
+to get the full and correct column name.
+
```
tracykteal: This is probably more confusing than helpful. We probably don't need to introduce this way of selecting. As noted this strategy can lead to issues and is probably not good practice.

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,143 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+>
+> 1. Create a `data.frame` (`surveys_200`) containing only the observations from row 200 of the
+>    `surveys` dataset.
+>
+> 2. Notice how `nrow()` gave you the number of rows in a `data.frame`?
+>
+>      * Use that number to pull out just that last row in the data frame
+>      * Compare that with what you see as the last row using `tail()` to make
+>        sure it's meeting expectations.
+>      * Pull out that last row using `nrow()` instead of row number
```
tracykteal: instead of 'the' row number

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,143 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+>      * Create a new data frame object (`surveys_last`) from that last row
+>      * Which
```
tracykteal: Is the 'Which' supposed to be here?

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,143 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
+### 4. Combine `nrow()` with the `-` notation above to reproduce the behavior of
+###    `head(surveys)` keeping just the first through 7th rows of the surveys
+###    dataset.
+
+```
+
+<!---
+```{r, purl=FALSE}
+## Answers
+surveys_1999 <- surveys[200, ]
```
tracykteal: should this now be 'surveys_200' instead of 'surveys_1999'?

02-starting-with-data.Rmd
```diff
@@ -111,10 +111,143 @@ Based on the output of `str(surveys)`, can you answer the following questions?
...
-As you can see, many of the columns consist of integers, however, the columns
-`species` and `sex` are of a special class called a `factor`. Before we learn
-more about the `data.frame` class, let's talk about factors. They are very
-useful but not necessarily intuitive, and therefore require some attention.
+## Inspecting `data.frame` Objects
+
+We already saw how the functions `head()` and `str()` can be useful to check the
+content and the structure of a `data.frame`. Here is a non-exhaustive list of
+functions to get a sense of the content/structure of the data.
+
```
tracykteal: suggest

'of the data.' -> 'of the data. Let's try them out!'

So that instructors know they should be typing them.

tracykteal:
Thanks for all the updates. Looks good and adds significant clarity.

PR #175, opened 2017-01-13
Revise intro

fmichonneau:


00-before-we-start.Rmd
```diff
@@ -89,25 +120,6 @@ workshop, you only need a `data/` folder.
...
-(top-left, in the default layout), the R console (bottom-left), your
-environment/history (top-right), and your files/plots/packages/help/viewer
-(bottom-right). The placement of these panes and their content can be customized
-(see menu, RStudio -> Preferences -> Pane Layout). One of the advantages of
-using RStudio is that all the information you need to write code is available in
-a single window. Additionally, with many shortcuts, autocompletion, and
-highlighting for the major file types you use while developing in R, RStudio
-will make typing easier and less error-prone.
-
-
```
tracykteal: Thanks, clearer without the additional information.

tracykteal:
Thanks, it makes more sense to introduce the definition of the syntax when we use them in the lesson. Showing that big script at the beginning tends to be more confusing than helpful.

PR #178, opened 2017-01-16
Add callout to avoid object/variable confusion

### LINKED ISSUE #108: Possible confusion about the term `object`
When working through the lesson material I noticed that what I thought of as `variables` was consistently called `objects`. Originally coming from a non `R` programming language I was slightly confused. So I did some research and it indeed seams to be more precise to use `object` instead of `variable` as the `R` language specification states:

> In every computer language variables provide a means of accessing the data stored in memory. R does not provide direct access to the computer’s memory but rather provides a number of specialized data structures we will refer to as objects. These objects are referred to through symbols or variables. In R, however, the symbols are themselves objects and can be manipulated in the same way as any other object. This is different from many other languages and has wide ranging effects. 

https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Objects
So as the wording is already correct and consistent throughout the whole lesson this is not really an issue. But there might be others like me that get confused so it might be a good idea to point out this special meaning of `object` or that they are sometimes also referred to as `variable` in the lesson material or in the instructor notes. This might be a possible pitfall for instructors as well as learners.

iimog:
Fix #108

### DIFF
--- a/01-intro-to-R.Rmd
+++ b/01-intro-to-R.Rmd
@@ -205,6 +205,10 @@ arguments.  If you don't, someone reading your code might have to look up the
 definition of a function with unfamiliar arguments to understand what you're
 doing.
 
+> Attention!
+> What is called `objects` in `R` is known as `variables` in many other programming languages.
+> Depending on the context `object` and `variable` can have drastically different meanings.
+> However, in this lesson the two words are used synonymously.
 
 ## Vectors and data types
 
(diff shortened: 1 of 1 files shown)


fmichonneau:
Could you please add a link to the R manual you list in the original issue?
Also, for the formatting, could you please:

- rename "Attention!" "Call out"
- use a level 3 header (use the ### prefix) for the title
- add a blank line between the header and the paragraph.

Thanks!

iimog:
Sorry for the extra work due to my wrong styling

tracykteal:
@iimog thanks for the contribution and it's no problem! That's why PRs are reviewed. It's just a part of the process.

iimog:
Thanks, Tracy

fmichonneau:
Thanks @iimog !

Issue #150, opened 2016-12-21
typo in dataframes page


mfoos:
Just above the "Inspecting data.frame Objects" header, there's a variable called "datarame" (twice) - should probably be changed to "dataframe" before a beginner tears out their hair debugging

[Could someone please tag this as beginner friendly?]

fmichonneau:
this challenge has been removed

PR #179, opened 2017-01-20
fixing-typo-in-r-ecology-lesson

gtlaflair:
I changed "datarame" to dataframe on handout-script.R. I believe this is related to issue #150.


tracykteal:
Thanks for noticing and fixing this typo @gtlaflair 

@fmichonneau is handout-script.R being generated from the Rmarkdown files? If so, do we need to fix it there instead of in handout-script.R?

fmichonneau:
Thanks for spotting this @gtlaflair.
It turns out we removed this challenge in the latest reorganization of the lesson.
@tracykteal is correct: the content of the handout-script is generated from the Rmd files.

PR #180, opened 2017-01-22
Lesson polishing

fmichonneau:
I went through the lessons to fix typos, grammar, and restructured the content a bit following the edits from last week.

The most significant change is merging the 2 chapters about data frames into a single one. Previously the split between the two was not really justified, and made things more confusing that they needed to be.

At some point, we also had a page just about the motivation for learning R. I removed it and merged its content with a smaller (less developed) section we had in the "before we start" chapter.

I think that once this is merged, we can announce the freeze, and get the lesson ready for release.


tracykteal:
Thanks, the combing and moving of text does add clarity.

PR #231, opened 2017-04-06
further specify use of filter - issue 225

straussed:
As suggested in issue 225, I further specified that filter is used for selecting rows based on certain  criteria. 

This is my first PR so let me know if I've done anything wrong!


fmichonneau:
It looks great! Thank you for contributing your first PR to data carpentry :)

Issue #218, opened 2017-03-17
Should we include Mac shortcuts?


dvanic:
The shortcuts in the lesson are for Windows (ctrl + enter). Should we also include the Mac ones (Cmd + enter) for learners/instructors?

fmichonneau:
I think that Crtl + Enter also works for mac, but it's a good idea to add them in the lesson. Any other OS-specificities may also need to be listed in the instructor guide.

ha0ye:
Working on it (will combine with other formatting fixes).

ashander:
awesome looks like this is closed by #235 

ps @ha0ye if you write 'closes #<number of the issue>' ([or some other similar keywords](https://help.github.com/articles/closing-issues-via-commit-messages/)) in the commit or PR it'll close the issue automatically when merged

ha0ye:
@ashander I did not know that. Thanks!

PR #235, opened 2017-04-07
formatting fixes and minor name fixes

ha0ye:
(see commit info for details, but addressing  ##182 and some other minor things I noticed)


ha0ye:
now also addressing #218

fmichonneau:
Thanks @ha0ye for your first contribution to Data Carpentry's R lesson!

These are really useful edits. I'm going to merge this PR, but I'm going to keep the R lower case in the file name of lesson 5. I think it's better to be consistent in using lower case everywhere, even if seeing a lower case r makes me cringe a little :grimacing:

ha0ye:
@fmichonneau, should the filenames for lesson 01 be lowercase then? It's currently "01-intro-to-R"...

fmichonneau:
Good catch! I'm going to rename it. Thanks for pointing this out

Issue #186, opened 2017-01-30
Check content of lesson for typos and grammar


fmichonneau:
Before release, we need to make sure that the lessons have correct grammar and no typos.

Ideally, we would need at least 2 people for each chapter.

- [ ] Before we start
- [ ] Introduction to R
- [ ] Starting with data
- [ ] Aggregating and analyzing data with dplyr
- [ ] Data visualization with ggplot2
- [ ] R and databases

achazhardenberg:
In the introduction to R lesson, at the very end, it says: "Now that we have learned how to write scripts, and the basics of R's data structures, we are ready to start working with the Portal dataset we have been using in the other lessons, and learn about data frames."

However, no dataset named "Portal" is mentioned in the previous lesson (Before we start).

kmpeck:
Working on this now.

fmichonneau:
There is now a new make target:

```bash
make check-spelling
```

that does spell checking on the lessons (Rmd and md files). There are no typos at this stage, but there might be some bad grammar left. I'll do another read over in a little while

PR #244, opened 2017-04-07
Fixed grammar and typos in 00-before-we-start.Rmd

kmpeck:
In reference to issue #186


tracykteal:
Thanks for these edits!

Issue #214, opened 2017-03-17
Rewording lesson objectives: before we start


dvanic:
_Describe the purpose of the RStudio script, console, environment, and plot windows._

Perhaps change to:

_Describe the purpose of RStudio's Source, Console, Environment, and Plots windows._

- this is what the windows are called in RStudio (at least on my Mac), so perhaps it's best if we use the same terminology as in Rstudio so as not to confuse leareners?

_Organize files and directories for a set of analysis in an R Project._

Perhaps change to:

_Organize files and directories for a set of analys**e**s in an R Project._

or 

_Organize files and directories for an analysis in an R Project._

fmichonneau:
good point. For the first one, maybe also replace "windows" with "panes". For the second one, maybe change "in an R project" with "as an R project".

tracykteal:
Addressed in https://github.com/datacarpentry/R-ecology-lesson/pull/248

PR #248, opened 2017-04-07
made text changes suggested in #214

emhart:
* Based one the suggestions of @fmichonneau and @dvanic I made textual changes to the before you begin

* Simplified the wording about dplyr.


tracykteal:
Thanks, nice updates.

Issue #222, opened 2017-03-17
Introduction to R: lesson objectives: working directory


dvanic:
Learning objective is:
Define the following terms as they relate to R: script, function, working directory, assign, object, variable.

Working directories are not discussed in this lesson at all, and are instead discussed in the "before we start" previous lesson. Should this objective be split into two, and placed in this and the previous lesson?

fmichonneau:
yes, splitting it to reflect the actual content of the lesson would be great

PR #250, opened 2017-04-07
moved the learning objectives around 

emhart:
Ensure that learning objective "working directory"  is in the right lesson to match content. Addresses #222


fmichonneau:
Thanks @emhart !

PR #257, opened 2017-04-19
Add template

ErinBecker:
@tracykteal

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,13 +2,8 @@ Please delete the text below before submitting your contribution.
 
 ---
 
-Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org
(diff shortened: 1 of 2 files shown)


fmichonneau:
LGTM. I'm wondering if we should add a sentence that would speak to people who are opening an issue/PR outside of instructor checkout. We routinely have contributors who are fixing typos while browsing the content for instance, and they might be put off by seeing this message.

ErinBecker:
@fmichonneau Not opposed in theory to changes, except that the same wording has been merged into a few other repos now and I think it would be good to have consistent templating language throughout if possible.

To my mind, the second para is general (all contributors) and the first directed toward people submitting contributions for instructor training.

fmichonneau:
OK, let's leave as it is for now, and if there is confusion we can revisit the wording. Thanks!

ErinBecker:
@fmichonneau Thanks!

PR #258, opened 2017-04-20
merge pull request templates

fmichonneau:
While cleaning up the repo, I realized that we recently merged a new template for pull requests, so I'm attempting here to merge the 2. It means that here the text for the PR template wouldn't be the same as in the other repositories. What do you (@ErinBecker @tracykteal) think?


ErinBecker:
@fmichonneau I like the check-list aspect of the previous template and I like the way you've unified them. My only concern is with the "I've read and followed the CONTRIBUTING document" and "I've used the correct formatting". While these are certainly things that we want newly contributed PRs to meet, I'm worried that this language may sound unfriendly and folks who didn't understand (or are worried they maybe didn't understand) something in the CONTRIBUTING document or the instructions for formatting will be hesitant to submit a PR. 

Could we maybe add a line to the effect of "If you're not about the formatting or have any questions, please contact XXX"? Or maybe "If you think the formatting might not be correct, please explain what part you're concerned about so the maintainers can take a closer look." 

I don't have the wording quite right yet, but what I"m going for is a single sentence that will signal to new contributors it's ok to make a mistake and their contribution is still welcome.

Thoughts?

fmichonneau:
Thanks for the feedback @ErinBecker! I agree with you. I edited the text slightly, but there might still be room for improvement.

Also @aurielfournier are you OK with having your email address listed in this pull request template?

PR #259, opened 2017-04-22
cjieming use r studio help interface

cjieming:
Add a screenshot and a section on seeking help using RStudio's built-in help interface


fmichonneau:
Thank you!

PR #264, opened 2017-05-10
added formatting dates in R

DanielleQuinn:
As per my discussion with @fmichonneau, I've added a brief demonstration of how {lubridate} can be used to convert date information from date components in separate columns to a POSIXct vector that can be added to a data frame as a new column called "date".

It was suggested that a plot incorporating date may be used to motivate this information; for the moment I've just included a brief how-to at the end of the Starting with Data segment of the lesson. I'd be more than happy to draft a complimentary piece for the Visualization segment, which could be separate or include the piece I've added here.

Looking forward to feedback and hoping to become more involved in SC and DC! : )

P.S. This is my first time creating a branch, modifying content, and creating a pull request - I apologize if its clunky or done improperly! : /

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -17,6 +17,7 @@ source("setup.R")
 > * Summarize the contents of a data frame in R.
 > * Manipulate categorical data in R.
 > * Change how character strings are handled in a data frame.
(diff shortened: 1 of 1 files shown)


fmichonneau:
It looks good! Thanks

Issue #95, opened 2016-06-10
A short intro to tidyr is needed


fmichonneau:
We need material that would introduce the basics of `tidyr`. It's a really useful package, and I think it needs to be introduced in the context of a data carpentry workshop

tracykteal:
Agree! @michaellevy has just been working on a tidyr lesson using the gapminder data and including the dplyr join. 

https://github.com/michaellevy/gapminder-R/blob/gh-pages/06-tidy-data.Rmd

I wonder if this could be adapted and brought in to these lessons.

fmichonneau:
Yes, that looks good! I like the idea of using gather() on a dplyr generated dataset.

I have been reluctant on adding joins from dplyr because there is not a lot of extra time, and we cover joins in the SQL lesson. It's an important concept though, so it might be worth seeing it twice over the course of a workshop.

jebyrnes:
I wonder if the dplyr lesson might incorporate this - a short two beat section on the end with gather and spread (not joins - that gets into its own thing) would go a long way, and doesn't feel like too much more on top of what is already there.

Is someone working on this? If not, I'm happy to add a bit using what's already there.

fmichonneau:
Hi @jebyrnes, no one is actively working on this as far as I know, so if you can create a short blurb about tidyr, that would be great.

tracykteal:
@michaellevy has added a nice section on TidyR for some social sciences lessons we've been trying. I think it would be great to incorporate in these lessons.

http://data-lessons.github.io/gapminder-R/06-tidy-data.html

https://github.com/data-lessons/gapminder-R
http://data-lessons.github.io/gapminder-R/

jebyrnes:
These are great! I've taken the approach of adding this to the dplyr section, labelling it more generally as data manipulation. I'll submit a pull request. I  realize this runs the problem of too much stuff in one lesson, but I've tried to minimize. I wonder, to some extent, if the data export section is too long? Perhaps we can cover `write.csv` in the previous lesson, but not have them go through the exercise of using dplyr to create the complete cases dataset?

PR #265, opened 2017-05-15
Add tidyr to dplyr lesson

jebyrnes:
So, this is an attempt at adding a short intro to `tidyr` as discussed [here](https://github.com/datacarpentry/R-ecology-lesson/issues/95) in the issues. I do worry this will add too much, but, feel very strongly given my own experiences as a quantitative ecologist that a working knowledge of gather and spread is essential.

I've tried to find the right place to slot it in, but I can see other options - separate small lesson and/or moving data writing to somewhere else.

Let me know what y'all think, but I do think this is important!


fmichonneau:
Thanks @jebyrnes, that's a useful edition.

PR #278, opened 2017-06-14
Update to RStudio new project dialogue text

benmarwick:
I'm using an RStudio daily, and I see "New project" rather than "Empty Project"


aurielfournier:
Great catch thank you!

PR #280, opened 2017-06-17
Added note on how to install missing package

mondorescue:
Suggested edit (as part of the instructor training checkout): I've added a blurb on how to install the `lubridate` package, if one gets an error prompt that it is missing.


aurielfournier:
Great thank you!

PR #281, opened 2017-06-20
cleaned up weight/hindfoot_length text in boxplot lesson and challenge

csjohns:


fmichonneau:
thanks, it looks great!

PR #282, opened 2017-06-23
Update 03-dplyr.Rmd

taylorreiter:
Add "s" to `survey`, making it `surveys`, to match the R object.


aurielfournier:
good catch thank you!!

PR #284, opened 2017-07-11
Adds a further resource about reprex

maelle:
Not sure whether I should have rather opened an issue. 😉 In any case Jenny's presentation about `reprex` and reproducible examples in general in that call was great so maybe useful to add.


fmichonneau:
Yes, definitely useful to add. Most of this section are resources for learners after attending a workshop, so it seems good to have here. Thanks!

PR #286, opened 2017-07-12
add data frame figure and accompying text to starting-with-data lesson

ab604:
This contribution is for instructor training, and I'll email to checkout@carpentries.org with a link to this contribution so you can record your progress.

My contribution is the creation of a figure to illustrate what a data frame looks like, with a minor addition to existing text to accompany the figure.

Thanks, 

Alistair ab604@soton.ac.uk


aurielfournier:
This is great Thanks!

PR #287, opened 2017-07-12
Update 03-dplyr.Rmd

maelle:
I added a link to the `tidyr` cheatsheet. RStudio now has one cheatsheet about `dplyr` only, and one about data import including `tidyr`. That said there's also the "old" one with both `dplyr` and `tidyr` at [this link](https://github.com/rstudio/cheatsheets/blob/master/source/pdfs/data-wrangling-cheatsheet.pdf) which you might prefer, although I'm not sure whether it's as updated as the two other ones.


fmichonneau:
thanks for adding this! I had missed the splitting of these cheatsheets!

maelle:
you're welcome! :smile_cat:

PR #289, opened 2017-07-18
Long data frame format described in preamble in ggplot lesson

GMoncrieff:

---
I added a sentence in the preamble about data frame format that ggplot likes. I have always found that if you understand that ggplot really prefers data in 'long' format, plotting becomes a breeze.
---


fmichonneau:
thanks, that looks like a useful addition!

Issue #285, opened 2017-07-12
SQL Databases feedback


JSurman:
I read through the SQL lesson and (as a relative novice to SQL) I think there needs to be a diagram or some more explanation of how the database tables work and what "Inner Join" and "Left Join" mean. Those concepts aren't introduced much and have some complexities/gotchas that might be worth explaining. I've seen some good diagrams on this since and if people think it would help I could try to make something? This could include how the databases are external to R.

Two other small things:
There's a challenge that optionally asks users to write an SQL query. I'm not sure what the point of this is? SQL query language hasn't been taught so a novice would just be basing their answer on the previous examples. It's optional so not a huge issue.
Also in the HTML version the results of the library(dplyr) command are displayed. It doesn't seem useful, is it intended?

aurielfournier:
Hello JSurman, Thanks for looking over the lesson. If you would be up for making such a diagram that would be a great help!

I will defer to @fmichonneau on the optional question. I am not sure why we have that particular one. 

While the output from library(dplyr) may not be that useful it is something that we want learners to become used to seeing, since they will encounter it when running their own code.

JSurman:
Righto, I'll have a stab at a diagram of the tables and joining/connections.

fmichonneau:
Hi @JSurman 

The lesson you refer to is supposed to be taught after the SQL lesson, so workshop participants would have already been exposed to writing SQL queries by this time. I do agree that it's not clear in the website (it's only mentioned in the instructor notes). 

I'm looking forward to seeing your diagrams!

Cheers

JSurman:
Do you think this would help? I just did the joins in this lesson, I believe this is what R does for SQL, it's what it does for frames...
[joins.svg.zip](https://github.com/datacarpentry/R-ecology-lesson/files/1154935/joins.svg.zip)

PR #290, opened 2017-07-20
Adding diagram to /img for SQL lesson

JSurman:
As discussed in issue #285, I have created a diagram for the SQL lesson to more fully explain how the different joins work. I renamed the tables Table_x and Table_y but it makes the join commands quite wide, if it looks OK we could leave it, or I could call the tables just 'x' and 'y'?


fmichonneau:
that looks great to me! the last thing would be the resize the page of the SVG file to match the size of the figure so we can include it directly into the lesson. Thanks for doing this!

JSurman:
I'm not really sure how to do that but I had a go, I think the pull request updated with that commit?

fmichonneau:
great, thanks! I'm going to add it into the lesson.

PR #293, opened 2017-07-28
Update to subsetting 02-starting-with-data.Rmd

KWierucka:
Changed order of examples explaining subsetting dataframes. The section now starts with explaining the simple concept of extracting a value based on its location in the dataframe, therefore showing that the first part of the command refers to the row and the second to the column. Then it moves on to subsetting whole columns in different ways.

This is a part of my instructor training checkout.


aurielfournier:
Great addition, thank you!

PR #297, opened 2017-08-01
update ggplot lesson to improve clarity

katrintirok:
I did some updates on the ggplot2 lesson to improve clarity, see for details below: 

- added information that ggplot2 is included in tidyverse
- added data load step in case someone had closed RStudio in between or had deleted workspace
- added a very short challenge before boxplot (maybe too simple?)
- removed `group` statement in aes() when `color` is introduced, since using `color` automatically groups data (removed all unnecessary `group` statements in later plots)
- adjusted text regarding the use of group/color accordingly
- removed colour=species_id from faceted plot
- added colour=sex in second faceted plot
- replaced panel.grid.major/minor.x/y with panel.grid (this way it is much shorter for the same result and less overwhelming)
- removed part where colour for sex is added (since that is now included earlier already)
- added note on width and height and font size when saving plots
- some minor typo/formatting corrections
- updated section with challenges under ### Data Visualization with ggplot2 in code-handout.R

### DIFF
--- a/03-dplyr.Rmd
+++ b/03-dplyr.Rmd
@@ -500,7 +500,7 @@ surveys_gw_wide %>%
 > 2. Now take that data frame, and make it long again, so each row is a unique
 >    `plot_id` `year` combination.
 >
-> 3. The `surveys` data set is note truly wide or long because both there are
+> 3. The `surveys` data set is not truly wide or long because there are
 >    two columns of measurement - `hindfoot_length` and `weight`.  This makes it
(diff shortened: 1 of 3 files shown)


fmichonneau:
Thanks for doing this @katrintirok, it's excellent. And I learned that grouping is now automatic :) 
The file `code-handout.R` is generated automatically from the content of the lesson, so the text needs to be modified in the lesson file directly. I'll take care of it though.

PR #301, opened 2017-08-16
added 'Arranging and exporting' section, which introduces gridExtra a…

justinmillar:
Combining and exporting ggplots was a major point of discussion in the recent R Data Carpentry workshop at the University of Florida. I added a section that introduces the gridExtra package for arranging plots in a single figure, as well as a brief explanation on why ggsave() is preferred over the Export options in RStudio.


aurielfournier:
This is slick thanks! I didn't realize you could save grid.arrange plots with ggsave.

Issue #268, opened 2017-05-17
Use readr to read and write csv files


fmichonneau:
Now that we have packages from the tidyverse available, I think we should consider swtiching to readr::read_csv and readr::write_csv. Thoughts?

kathy0305:
Yes, it only makes sense to do so; we need to be consistent.
Tasks:

- Convert all files/ lesson that contain read.csv or write.csv to read_csv and write_csv.
- Lesson "Starting with Data": What are Data Frames, include tibbles explanation
- Lesson "Starting with Data": stringsAsFactors=FALSE lesson. Delete/update

Questions: 
- Should we still refer to them as  Data Frames or as tibbles ?? tibbles are data frames, so maybe keep the data frame references??
- Should we tackle read_table??

aurielfournier:
I am fine with that. But I think we would atleast still need some kind of mention/discussion of read/write.csv. Since learners will see it all the time in examples/others code, and I would want them to be aware off it, and esp of the stringsAsFactors. 

I'm a heavy tidyverse user, but I still use read.csv. so maybe I'm just being a grumpy late adopter.

kathy0305:
Yep I agree 100%. I struggled with that idea myself when I designed my workshops that are based on tidyverse, the workshops is for R Newbies. I didnt want my students to come back to me and ask about $ and read.csv ...and base R stuff. So I have a base R introductions too.

WeeBeasties:
Is anyone currently working on this? I'm completing instructor training and this looks like a way that I could make a meaningful contribution. I currently use both base R and the tidyverse to load data into my projects. So, I'm comfortable enough take it on.

kathy0305:
I am working on updating it! my apologies for not logging the issues here. Francois and I have been communicating through email, to figure out the best way to do this.  We came to a consensus last week, and I'll be updating them this week.

WeeBeasties:
I kind of assumed so (but I try to assume very little in life). I definitely didn't want to duplicate effort or step on toes. There's plenty to be done; I'll find another way to chip in. Cheers!

kathy0305:
I am in the same boat, wanted to make sure I don't go overboard with the changes and wanted to get the green light before I assume things.  I actually have the changes ready to go, just tweaking few things. I'll appreciate another pair of eyes after that .

PR #303, opened 2017-08-22
changed read.csv to read_csv (tidyverse way)

kathy0305:
changed read.csv to read_csv , since we are teaching with tidyverse pkg, we should stick with tidyverse syntax and create tibbles

added library(hexbin) after installation of pkg.

added () after tally function

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -7,7 +7,7 @@ minutes: 60
 
 ```{r setup, echo=FALSE, purl=FALSE}
 source("setup.R")
-surveys_complete <- read.csv(file = "data_output/surveys_complete.csv")
+surveys_complete <- read_csv(file = "data_output/surveys_complete.csv")
 ```
 
 ```{r, echo=FALSE, purl=TRUE}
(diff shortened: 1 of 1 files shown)


fmichonneau:
Hi @kathy0305 

Thanks that is very useful and something that we need to do for our lessons (see #268). We need to be consistent though and if we change it in the ggplot lesson, we need to change it everywhere and possibly rework the part where we first introduce `read.csv` and the activity with using `stingsAsFactors` in Chapter 2. We also need to change it where we use `write.csv` to replace it with `write_csv`. Is it something you would be interested in tackling?

kathy0305:
I meant to make just the small changes first, by adding () and the library.....then saw the read.csv and made that change. I should have looked at all the lessons, my apologies. Yes I would like to tackle all the read.csv and write.csv to read_csv and write_csv. I think that we need to be consistent with the tidyverse way, besides its faster.  
Let me look at chapter 2 and see how can we revise it.

kathy0305:
Ok, I just looked at chapter 2 , where it talks about the need to include stringsAsFactors=FALSE, when reading a file. That should be an easy task to tackle, as read_csv does not convert char to factors, so we simply delete that section. Since we are teaching to students/people not familiar with R, I don't think there is a reason to further explain. But perhaps add something about tibbles and what happens to column names. Let me know!

Issue #274, opened 2017-05-28
description / definition of atomic vector


loriling96:
for lesson "01-intro-to-r", I think there is potential confusion about what IS an atomic vector for a learner. Atomic vector is first defined as a data type in R, of which there are six sub-types (line 307).  However, later in the lesson in the example for Missing Data, the comment on the output of the function "na.omit(heights)" is that "the returned object is atomic." (line 497). This wording is potentially confusing and suggests that atomic is something distinct from numeric. The returned object of na.omit(heights) is not a simple numeric vector and perhaps more description is needed.

PR #304, opened 2017-08-24
Atomic vector explanations in 01-intro-to-r.Rmd

kmpeck:
In response to issue #274 , I clarified the explanation of an atomic vector when it is first encountered in the lesson. I also added in text to let the reader know that numeric and double are synonymous terms. This is important because when checking the type of a vector in R, it outputs "double" which could be confusing if a reader has never encountered the term. I also added in notes for the function typeof() for checking the type of a vector. Further, I fixed the text confusion under the examples in the missing data section, changing it from reading "the output is atomic" to "the output is an atomic vector of type numeric (or double)".


aurielfournier:
This is really great @kmpeck thank you so much !

PR #305, opened 2017-08-31
Amend spread and gather dplyr lesson

ab604:
I've made a fairly substantial change to the spread and gather section of the lesson. Hopefully for the better. 
I was attempting to clarify the definition of tidy data and the motivations for the transformations, rather than discussing in terms of 'long' and 'wide'. I've added two figures to help illustrate what the spread and gather functions do and how they work.


fmichonneau:
thanks for this! I'll look into it in more details tomorrow.

fmichonneau:
You put a lot of work into this and it's much appreciated. I really like the figures you made, it makes `gather` and `spread` more intuitive. This is however too long given our time constraints for the workshop, and I can't merge it as is. Here are a few suggestions to cut down the length:

- the description of tidy data happens during the spreadsheet lesson and do not need to be repeated here
- personally, I find that motivating the use of tidyr on an already tidy dataset by saying "tidyness" is context dependent is confusing. To me it seems that it's better to motivate tidyr here in the context of wanting to present the data differently to make a table for a publication for instance. Then it makes sense to want to have one column for each genus. We may also want to mention that for types of visualization it is useful to be able to go from long to wide format.

We don't talk about `gilmpse()` anywhere else in the lesson, and I am in favor of not including it. I don't think it is that different from `str()` and it would increase the cognitive load to our learners.

If you could modify your PR to have one paragraph to motivate tidyr, and keep the sections about gather and spread as they are, that would be great.

ab604:
Thanks Francois, this is really helpful. I’ll make the changes and resubmit a pull request.

Earlier this year I did actually fork the lesson and do a major re-write of the whole thing if you’d like to take a look and see if there are other aspects you might think are worth merging (with further changes) into the existing lesson. I can send you the link if you are interested? I suspect the figures will be most appropriate. I had actually dropped gather and spread, ironically due to cognitive load concerns, hence I’d come back to it here. On the point about wide and long for visualisation, I was trying to move towards the point that Hadley makes in his paper that one needs to think in terms of variables and observations, and when we turn columns into rows we are converting variable names into values of a new variable and observational unit. That it changes the shape is a side effect, not the aim itself. With rectangular data, one plots values of variables corresponding with observations, the tricky part is understanding which is which – and that is determined by the data science question one is trying to answer. What’s really confusing (to me at least) is that prior to spreading the data (in the first example used), our previously tidy data would be considered untidy, but only because the question has shifted from one about individual rodent records to one about average species weight. I haven’t really come up with a better way to explain this, but I’ll think some more. It speaks to the problem of dataset familiarity which will always be a problem for workshop attendees. Not only do they have to take on the new tools, but the data is unfamiliar too.

There’s a lot I’d change in my forked re-write, but the motivation was to make the lesson more problem orientated. When I first encountered the ecology lesson I was in the position of a workshop attendee and I didn’t really understand what the data was or why the people who gathered it had done so. And therefore the tools and manipulations didn’t seem quite as tangible as when there is a specific question that would require one, to for example, filter, select or plot the data. So I spent some time reading about the Portal project and created a version of the lesson with the aim of demonstrating reproducible science using one of the papers, whilst remaining true to the original lesson, and signposting the cross over with SQL and the spreadsheet lesson.

I’ve had a fair amount of discussion with colleagues here about the cognitive load aspect and the pros and cons about mixing base R with tidyverse, and whether we should be teaching just one or the other. Personally, I started using R in 2014 and came into it from the tidyverse side of things through the work of the John Hopkins team, Hadley Wickham and Garrett Golemund, and hence I like their approach. But one of the things I like about R is the rich community and the different approaches to the same problems. I probably just defaulted to glimpse() as that’s what I usually use!

Likewise I had discussions about the cognitive load issues around using different data sets that are simpler to understand for certain examples than the ecology data, such as the mtcars or diamonds datasets. I personally like the ability to mix simple and more complex examples, but in the workshops I’ve attended thus far, only a fraction of the material ever gets covered, so I’m leaning towards the idea of making different versions of the lessons at the moment. Lots of attendees ask for more advanced stuff, so it’d be nice to have a more stripped down fundamentals version, and perhaps a more advanced one that delves into simple modelling (i.e. linear regression) and report generation.

All the best,

Alistair

ab604:
Hi again,

I've revised things now, hopefully this is better. 

Further to my last comments and your feedback, I've tried to make it more concise and clarify the motivation in terms of making new tables that reshape the data according to observations of interest. Hopefully this is easier to follow than talking about tidyness being context dependent. 

Additionally I've cross referenced the spreadsheet lesson and changed `glimpse()` to `str()`.

fmichonneau:
Thanks for you thoughts. You make some very good points, and I have started to work on a rewrite of this lesson that would be more problem oriented. I would like to see what you did though, and if you have any other figures and visual aids that we could use in this lesson, they would be most welcome.

There are other data carpentry lessons being developed, including one that uses the gapminder dataset that is probably easier to grasp for workshop participants who are not ecologists by training. 

There has been some discussions about including some basic statistics in the lessons, but in most cases, it's beyond the scope of our workshops.

Your latest changes look good. I'll merge them, and re-generate the site soon.

PR #320, opened 2017-10-27
Consistent styling

shiva1387:
To be consistent with other chapters in the lesson where shortcuts are mentioned for both Windows and Mac OS, added a shortcut for mac for the <- operator for mac in 01-intro-to-r lesson.

A more pedantic issue, specifically mentioned that: "Columns of class character are never converted into factors automatically while loading the dataset."


fmichonneau:
Thank you for your contribution. I'm not going to merge this right away, as I need to think a little more about how and where to deal with `read_csv` in general in the lesson. We are still using a mix of `read_csv` and `read.csv` and it needs to harmonized everywhere.

PR #322, opened 2017-11-03
change read.csv to read_csv in lesson 5 (R and Databases)

ChuckFinley:
There are a few outstanding issues in the R and Ecology curriculum noting inconsistencies in usage of read.csv (from base) and read_csv (from readr). read_csv and the readr package are introduced midway through Lesson 3 dplyr. This commit changes read.csv to read_csv in Lesson 5 R and Databases. There are remaining read.csv calls in Lesson 2 Starting with Data and Lesson 3 dplyr, but since these come before the introduction of read_csv they’ll require a more thorough restructuring.


fmichonneau:
thanks for this!

Issue #327, opened 2017-11-19
Selecting/filtering improvement: Demo *why* pipes are so helpful


jenniferthompson:
It might be helpful to have short examples of intermediate and nested methods to compare to a pipeline, like the below, so that students can better understand the benefit of the pipe. (Happy to open this as a PR if others agree.)

```
library(tidyverse)

surveys <- read_csv("data/portal_data_joined.csv")

## Example of intermediate steps
surveys2 <- filter(surveys, weight < 5)
surveys_sml <- select(surveys2, species_id, sex, weight)
## Naming things is hard (how to keep track, esp with >2 steps?); workspace is cluttered

## Example of nested functions
surveys_sml <- select(filter(surveys, weight < 5), species_id, sex, weight)
## No need to name a bunch of things, but hard to read

surveys_sml <- surveys %>%
  filter(weight < 5) %>%
  select(species_id, sex, weight)
## Most readable, no need to name a bunch of things, yay
```

fmichonneau:
hi @jenniferthompson  -- Yes a short PR would be great!

jenniferthompson:
Sounds great @fmichonneau! I'll work on it this week.

aurielfournier:
I agree an example would be great. This is something I've struggled to teach on the fly several times, would be great to have as apart of the lesson

PR #328, opened 2017-11-21
Add examples of intermediate steps + nested functions

jenniferthompson:
I added examples of intermediate steps and nested functions, as discussed in issue #327. (I also modified text to clarify that `read_csv` is from the `readr` package (not dplyr) - since that wasn't discussed, I can remove it if needed!)


fmichonneau:
thanks for your contribution! it looks great.

PR #341, opened 2017-12-27
Bbolker pipe parens

bbolker:
This is an *optional* pull request. My co-instructor feels strongly, and I feel moderately strongly, that the syntactic sugar of functions without parens in pipelines (e.g. `%>% tally` or `%>% head`), is not worth the extra cognitive load/weirdness.

Arguments for:

- fewer characters to type and read
- learners might see it in existing code bases

Arguments against:

- it's an extra piece. We already have the "functions in pipelines have their first argument inserted automatically" concept established, we don't need the extra shortcut.


aurielfournier:
I agree @bbolker, what are your thoughts @fmichonneau ?

fmichonneau:
I agree as well. Thanks @bbolker!

PR #342, opened 2017-12-27
Bbolker gatherfirst

bbolker:
This is the biggest/most significant pull request.

I like the examples given previously and have retained them, but in my experience **the** most common use case for `tidyr` is when I'm given a (relatively) wide data set and need to make it long in order to work with tidyverse etc. (also most statistical machinery in R).

I made up a new data set (machinery is in this pull request) based on species counts; this would need to be stored somewhere more sensible/permanent.

I also grabbed an image from the SC tidyverse lesson.

PS: the "four rules" described at the top of this section don't actually show up in the spreadsheet lesson (maybe they used to?) They are in Hadley's "tidy data" paper.


fmichonneau:
Several people have already mentioned they would like to review the concept of tidy data in the context of the R lesson. In parallel, I agree that the concept of tidy data needs to be reinforced in the spreadsheet lesson (especially if the 4 rules aren't included in the lessons anymore). I think a lot of what is included here is needed, and could take the place of a revamped "intro to data frames" now that we are focusing more on the tidyverse. We'll be working on an overhaul on the lessons and a new release in the next few weeks and we will incorporate these changes then.

PR #343, opened 2017-12-27
style/grammar fixes

bbolker:
The only substantive/subjective points:

- a bean plot is not quite the same as a violin plot (see [here](https://cran.opencpu.org/web/packages/beanplot/vignettes/beanplot.pdf)); it also includes horizontal segments for the observed data points.
- I think the second argument of a ggplot is technically a "mapping", not an "aesthetic" [actually I guess it's an "aesthetic mapping": http://ggplot2.tidyverse.org/reference/aes.html]


aurielfournier:
this is great thanks @bbolker

Issue #344, opened 2018-01-03
no warning message for lubridate parse failure (starting-with-data)


elahi:
After the following command in the lubridate section:

``` {r}
surveys$date <- ymd(paste(surveys$year, surveys$month, surveys$day, sep = "-"))
```

The text states "Did you notice the warning message?".  Not sure if this is a problem on my machine, but I did not get a warning message (after re-installing lubridate, still no warning).  If others also notice this discrepancy, it may be worth a small change in the text.  

In addition, it may also help to build up the lubridate example a bit.  For example:

```

my_date <- ymd("2015-01-01")
str(my_date)

my_date <- ymd(paste("2015", "1", "1", sep = "-"))
str(my_date)

# What happens when we try to define an invalid date?
ymd("2017-02-30", quiet = T) # no warning, even though quiet is TRUE
?ymd

# Now let's create a date in our surveys dataset
surveys$date <- ymd(paste(surveys$year, surveys$month, surveys$day, sep = "-"))
str(surveys)

# Check for NAs 
summary(surveys$date) # yes, some dates failed to parse

# Create a character vector of our dates
surveys_dates <- paste(surveys$year, surveys$month, surveys$day, sep = "-")

surveys_dates[is.na(surveys$date)]
```

Let me know if this is worthy of a pull request.

aurielfournier:
Hi @elahi you are right that the warning message is not longer there, ymd is just putting in the NAs without warning us now. 

I think the example you have presented is good, and would welcome a pull request

please use `TRUE` instead of `T` though, as our lessons encourage learners to write out the full words to avoid issues. 

Thanks!

elahi:
Done, thanks for the quick response @aurielfournier .

PR #345, opened 2018-01-03
modified text to reflect no warning message from lubridate parse failure

elahi:
This pull request is in reference to issue #344. 
Please edit as you see fit.  
thanks - Robin

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -555,7 +555,25 @@ Start by loading the required package:
 library(lubridate)
 ```
 
-Create a character vector from the `year`, `month`, and `day` columns of
+Let's create a date object and inspect the structure:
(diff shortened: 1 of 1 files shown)


aurielfournier:
Great thank you @elahi !

plantarum:
> What happens when we try to define an invalid date? It is worth inspecting the help page for ymd().

````
ymd("2015-02-30", quiet = TRUE) # No warning, even though quiet is TRUE
?ymd # Note that the default value of quiet is TRUE, which should return a warning if a date failed to parse
````

I think this isn't correct. The default value of quiet is `FALSE`, not `TRUE`, and when it's `FALSE` it does issue warnings. So in the example above you've deliberately turned off warnings, then suggested that you should have got one anyways? Unless something has changed in lubridate recently. I just installed version 1.60.

PR #346, opened 2018-01-03
adding in emphasis that things be installed locally

aurielfournier:
At my last workshop we had extensive issues (more then half of computers) with problems installing R/R packages. Unfortunately these issues were not realized until we were into the first days lessons, and again at the start of day 2. I don't have a solution for this, but wanted to add some text to let instructors know to be on the look out for this and ask about it if a learner is having issues installing an R package/R itself.


aurielfournier:
This is in relation to #346

Issue #336, opened 2017-12-20
improve setup instructions for networked machines to ensure local install


aurielfournier:
I just taught this lesson set, and we had large scale issues with non-local R installs and R packages. Need to spend some time working on improving the set up instructions and documentation to help folk who have computers which are networked install things properly.

PR #350, opened 2018-01-03
adding link to necessary_packages.R

aurielfournier:
adding link to instructor guidelines so that they can find the list of all the necessary packages.


aurielfournier:
Also in reference to #336

PR #351, opened 2018-01-16
Added a sentence on how the assignment operator can also face an opposite direction

atwgao:


fmichonneau:
hi @atwgao 

Thank you for your contribution. After discussing with the other maintainers, we think that showing a third way of doing assignments will be confusing to new learners, especially given that the forward assignment is generally not recommended in the various style guides. 

I also noted you updated the YAML to change the style of the tables. Is there one in the lesson that would benefit from this change? If so, could you please create a pull request against `_site.yml` that stores the output options for all pages?

PR #354, opened 2018-01-22
fix confusion around parsing invalid dates

fmichonneau:
@aurielfournier  could you please review this?

I suggest we change the narrative a little so we don't rely on the bug currently affecting `ymd`. We use the function, check it produces the correct output, and investigate why it doesn't. It seems like a realistic approach that is good to demonstrate and teach. Let me know what you think.

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -556,67 +556,56 @@ library(lubridate)
 ```
 
 Let's create a date object and inspect the structure:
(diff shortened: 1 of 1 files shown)


aurielfournier:
I will look this over by the end of the week, the federal shutdown here has really messed up my week, and even if they do come back tomorrow things are still going to be bumpy for a few days

aurielfournier:
@fmichonneau I think this is great, focuses more on the learning, instead of the bug. Nice work!

PR #361, opened 2018-02-06
Update 04-visualization-ggplot2.Rmd

hrahman2:
When introducing ggplot, I think it is good to begin with the general structure/template of how to use the function, such as:
 
ggplot(data = <DATA>) + 
  <GEOM_FUNCTION>(mapping = aes(<MAPPINGS>))

This way, throughout the remainder of the lesson, instructors can return to this template when introducing different types of functions (geom_point, geom_bar, etc) and students will have this template in mind if they encounter a geom_function we do not cover.

I added this to the beginning of the section please, but please not other sections may have to be updated if this change is made.

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -55,7 +55,9 @@ when making figures with ggplot.
 ggplot graphics are built step by step by adding new elements. Adding layers in
 this fashion allows for extensive flexibility and customization of plots.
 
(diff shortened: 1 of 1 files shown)


fmichonneau:
thanks @hrahman2, this looks good.

Could you please change where the `aes()` goes? in almost all the code below, we put it inside the `ggplot()` call. We could talk about the difference between the two here, but it feels very early to introduce this type of complexity.

PR #365, opened 2018-02-26
changed variables for objects to avoid confusions

scortijo:
Hello,

In the intro to R, I changed 'variable' for 'object' to avoid confusion. The two were used interchangeably quite early in the lesson (from the objectives), before the explanation about 'objects vs. variables' 

Sandra Cortijo
---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

### DIFF
--- a/01-intro-to-r.Rmd
+++ b/01-intro-to-r.Rmd
@@ -15,7 +15,7 @@ source("setup.R")
 >
 > * Define the following terms as they relate to R: object, assign, call,
 >   function, arguments, options.
-> * Assign values to variables in R.
+> * Assign values to objects in R.
 > * Learn how to _name_ objects
 > * Use comments to inform script.
 > * Solve simple arithmetic operations in R.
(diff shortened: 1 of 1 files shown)


mondorescue:
Thanks @scortijo for catching this. I think your changes are reasonable.

fmichonneau:
Thanks for this @scortijo 

I like the idea of being consistent in the choice of our words. However, in other episodes we refer to both words. We tend to treat them as synonyms. To me (and that maybe an inaccurate personal interpretation), I tend to use "variable" when I talk about the name of what is stored in memory, and "object" when I talk about what is actually stored in memory, but the distinction of the two in a sentence can be blurry.

What do others think?

fmichonneau:
As a reminder, we have a note in our lesson saying we treat both as synonyms.
This https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Objects explains how R defines objects.

scortijo:
Hi François,

I know there is a note about variable and objects (which is great). Perhaps it can be moved earlier in the lesson? I guess after that note, it's nice to use both variable and object. But before the note, we might try to be consistent?


________________________________

mondorescue:
@fmichonneau I agree with your interpretation of "variable" vs "object". I do think it can be confusing to someone with little or no programming / computer science background. We should consider moving the *Objects vs. variables* section to the top as suggested by @scortijo , put it in a call-out box after the first paragraph in the *Creating objects in R* section, or maybe change this sentence from

> However, to do useful and interesting things, we need to assign values to objects.

to

> However, to do useful and interesting things, we need to assign values to objects (or variables).

which should imply that the term objects and variables are interchangeable; and later in the *Objects vs. variables* subsection, the reader gets the full story.

fmichonneau:
@mondorescue your suggestions make sense to me. @scortijo would you be willing to implement this?

PR #372, opened 2018-03-06
Update the note section 

Arsh-07:
Add some comments on the use of read.table and how similar result can be achieved as with the read.csv function.


aurielfournier:
@Arsh-07 thanks for adding this detail. This is a good thing to point out to learners, since of course they won't always be working with .csv files.

PR #379, opened 2018-03-10
Added elaboration to placing a vector as an input to color argument in ggplot

ShubhangG:
There was no elaboration for adding a vector as an input for the color argument in ggplot, I added a further explanation that may help.


aurielfournier:
Thanks @ShubhangG this is a great addition.

PR #382, opened 2018-03-13
Import data from Import Dataset > From Text (readr) ...

maurolepore:
RStudio provides a GUI interface to import datasets. This, combined with the __readr__ package solves a number of common and important problems. 

* Reduces frustration because most students are familiar with GUI and not with typing paths.
* Avoids converting strings to factors -- a topic that is too early to address. `readr::read_csv()` defaults to `stringsAsFactors = TRUE`.
* Introduces students to `readr::read_csv()` ASAP. Some cool features the students may want to know about include:
    * it lets user specify what to interpret as `NA`, 
    * it lets user specify column types. 

I show an example [here](https://fgeo.netlify.com/2018/03/13/2018-03-13-import-dataset/).

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -58,10 +58,19 @@ download.file("https://ndownloader.figshare.com/files/2292169",
               "data/portal_data_joined.csv")
 ```
 
-You are now ready to load the data:
+You are now ready to load the data. Importing a dataset into R can be challenging. It often results in silent issues that cause severe errors later in your data analysis. Getting this right is important, and also it is simple if you use RStudio and the __readr__ package.
+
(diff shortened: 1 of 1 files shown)


aurielfournier:
Hello @maurolepore thank you for the suggestion. 

We appreciate your input on how to reduce frustration for our learners in regard to importing data. This is often one of the most frustrating parts of working with R. Because of that though, one of our goals is to help walk learners through that process with the code, not bypass it entirely. We want to spend some time talking about file paths and working directories so that learners are ready for applying what they learn in our workshops to their own tasks. 

So for now we will not be including this in the lesson, but we appreciate your thoughtfulness.

Issue #368, opened 2018-03-01
Change order in "Getting set up" in Before we start


fmichonneau:
As mentioned in #366, it might be better to download the script before creating the "data" folder.

Amrithasuresh:
@fmichonneau Simply changing the order will do? Like below. 

- Download the code handout, place it in your working directory and rename it (e.g., data-carpentry-script.R).
- Under the Files tab on the right of the screen, click on New Folder and create a folder named data within your newly created working directory (e.g., ~/data-carpentry/data)

fmichonneau:
yes exactly. Please let us know if you need help creating your pull request. Thanks!

PR #383, opened 2018-03-14
Change order in "Getting set up" in Before we start #368

Amrithasuresh:
I have updated accordingly. Thank you. #368


fmichonneau:
thanks!

PR #386, opened 2018-03-18
Correct mistakes and suggest wording changes

prosoitos:
Correct a few mistakes, in particular:
- loading the tidyverse does not load all the packages. Lubridate, for instance, does not get loaded by `library(tidyverse)`
- unnecessary combine function `c` in the indexing section
- several small grammar or syntax mistakes

Suggest wording changes:
- cut unnecessary words

One problem not addressed (I will open an issue as I don't know rmarkdown enough to address it myself-I use org mode, not rmarkdown in my own projects):
- text answer appears at the bottom of the question in the last challenge

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -97,7 +97,7 @@ functions `read.csv()` or `read.table()`; in other words, when importing
 spreadsheets from your hard drive (or the web).
 
 A data frame is the representation of data in the format of a table where the
-columns are vectors that all have the same length. Because the column are
+columns are vectors that all have the same length. Because columns are
 vectors, they all contain the same class of data (e.g., characters, integers,
 factors). For example, here is a figure depicting a data frame comprising a
 numeric, a character, and a logical vector.
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -200,7 +200,7 @@ surveys[, 1]    # first column of the data frame (as a vector)
 surveys[1]      # first column of the data frame (as a data.frame)
 surveys[1:3, 7] # first three elements in the 7th column (as a vector)
 surveys[3, ]    # the 3rd row of the data frame (as a data.frame)
-surveys[1:6, ]  # the first 6 rows of the data frame. This is equivalent to head(surveys)
+head_surveys <- surveys[1:6, ] # equivalent to head_surveys <- head(surveys)
 ```
 
 `:` is a special function that creates numeric vectors of integers in increasing
@@ -209,12 +209,11 @@ or decreasing order, test `1:10` and `10:1` for instance.
 You can also exclude certain indices of a data frame using the "`-`" sign:
 
 ```{r, purl=FALSE}
-surveys[,-1]          # The whole data frame, except the first column
-surveys[-c(7:34786),] # Equivalent to head(surveys)
+surveys[, -1]          # The whole data frame, except the first column
+surveys[-c(7:34786), ] # Equivalent to head(surveys)
 ```
 
-As well as using numeric values to call indices of a `data.frame` (or `matrix`), columns
-can be called by the names of the variables they contain, using one of the four following notations:
+Data frames can be subset by calling indices (as shown previously), but also by calling their column names directly:
 
 ```{r, eval = FALSE, purl=FALSE}
 surveys["species_id"]       # Result is a data.frame
@@ -293,9 +292,7 @@ of a special class called `factor`. Factors are very useful and actually
 contribute to making R particularly well suited to working with data. So we are
 going to spend a little time introducing them.
 
-Factors represent categorical data. They are stored as integers associated with labels. While factors look (and often behave) like character vectors,
-they are actually integers under the hood. So you need to be very careful when
-treating them as strings.
+Factors represent categorical data. They are stored as integers associated with labels and they can be ordered or unordered. While factors look (and often behave) like character vectors, they are actually integer vectors under the hood. So you need to be very careful when treating them as strings.
 
 Once created, factors can only contain a pre-defined set of values, known as
 *levels*. By default, R always sorts *levels* in alphabetical order. For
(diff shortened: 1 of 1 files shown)


02-starting-with-data.Rmd
```diff
@@ -290,19 +288,14 @@ feature to get the full and correct column name.
...
 
-Factors are used to represent categorical data. Factors can be ordered or
-unordered, and understanding them is necessary for statistical analysis and for
-plotting.
-
-Factors are stored as integers, and have labels (text) associated with these
-unique integers. While factors look (and often behave) like character vectors,
-they are actually integers under the hood, and you need to be careful when
-treating them like strings.
+Factors represent categorical data. They are stored as integers associated with labels. While factors look (and often behave) like character vectors,
```
fmichonneau: Why do you suggest removing the part about ordered and unordered factors?

02-starting-with-data.Rmd
```diff
@@ -532,19 +525,19 @@ converting date and time information into a variable that is appropriate and
...
+We are going to use the `ymd()` function from the package **`lubridate`**. **`lubridate`** belongs to the **`tidyverse`**-a package containing a group of packages that work well together. You can install **`lubridate`** directly (`install.packages("lubridate")`)) or by installing the **`tidyverse`** (`install.packages("tidyverse")`)). When you load  the **`tidyverse`** (`library(tidyverse)`), the chore packages (the packages used in most data analyses) get loaded. **`lubridate`** however does not belong to the chore tidyverse, so you have to load it explicitly with `library(lubridate)`.
```
fmichonneau: chore -> core

02-starting-with-data.Rmd
```diff
@@ -421,25 +414,22 @@ head(sex)
...
-your data, and converting as a factor only the columns that require this data
+In most cases, it is preferable to set `stringsAsFactors = FALSE` when importing
+data and to convert as a factor only the columns that require this data
 type.
 
 Compare the output of `str(surveys)` when setting `stringsAsFactors = TRUE`
 (default) and `stringsAsFactors = FALSE`:
 
 
 ```{r, eval=FALSE, purl=FALSE}
```
fmichonneau: There is a little issue with the formatting here. Could you please delete the "Compare the output..." sentence from the text, bring back the sentences deleted from the chunk, and switch to `purl=TRUE`? this way the code handout will include this chunk.

fmichonneau:
Could another maintainer please review this PR as well, and request any other updates in case i missed anything before @prosoitos starts working on the changes I requested?

02-starting-with-data.Rmd
```diff
@@ -196,25 +196,25 @@ different classes.
...
 surveys[1, 6]   # first element in the 6th column (as a vector)
-surveys[, 1]    # first column in the data frame (as a vector)
-surveys[1]      # first column in the data frame (as a data.frame)
+surveys[, 1]    # first column of the data frame (as a vector)
+surveys[1]      # first column of the data frame (as a data.frame)
 surveys[1:3, 7] # first three elements in the 7th column (as a vector)
-surveys[3, ]    # the 3rd element for all columns (as a data.frame)
-head_surveys <- surveys[1:6, ] # equivalent to head_surveys <- head(surveys)
+surveys[3, ]    # the 3rd row of the data frame (as a data.frame)
+surveys[1:6, ]  # the first 6 rows of the data frame. This is equivalent to head(surveys)
```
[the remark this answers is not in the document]
prosoitos: Will do. I was trying to be consistent with the rest of the examples and I didn't realize that this was also teaching something else.

02-starting-with-data.Rmd
```diff
@@ -196,25 +196,25 @@ different classes.
...
+You can also exclude certain indices of a data frame using the "`-`" sign:
 
 ```{r, purl=FALSE}
 surveys[,-1]          # The whole data frame, except the first column
 surveys[-c(7:34786),] # Equivalent to head(surveys)
 ```
 
-As well as using numeric values to subset a `data.frame` (or `matrix`), columns
-can be called by name, using one of the four following notations:
+As well as using numeric values to call indices of a `data.frame` (or `matrix`), columns
```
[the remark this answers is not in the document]
prosoitos: I wanted to explain somewhere what those numbers were (they represent the indices of the df) as I felt that this wasn't explicit anywhere. And the idea of subsetting is true as much for the use of numeric value as it is for the use of column names and I can see that this is what you were trying to explain, but I felt that the sentence wasn't making this clear. Maybe a better approach would be along the line: "Data frames can be subset by calling indices (as shown previously), but also by calling column names directly:"

Let me know if you would like me to try something like that.

02-starting-with-data.Rmd
```diff
@@ -255,7 +253,7 @@ feature to get the full and correct column name.
 > ## 3.
 > surveys_middle <- surveys[nrow(surveys)/2, ]
 > ## 4.
-> surveys_head <- surveys[-c(7:nrow(surveys)),]
+> surveys_head <- surveys[-(7:nrow(surveys)), ]
```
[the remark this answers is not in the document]
prosoitos: What is the notation that gets added? The use of ":" to create vectors of numeric values by increment of 1 already appears in the lesson (e.g. in "surveys[1:3, 7]"). 

I think that the use of `c` here is improper (even if it works) since `7:nrow(surveys)` is already a vector and the function of `c` is to form a vector from elements. So it is redundant and unnecessary (and not a really good practice).

Would you find `surveys_head <- surveys[-7:-nrow(surveys), ]` an easier form for the students?

02-starting-with-data.Rmd
prosoitos: I was trying to declutter this section a little. To keep the idea of odered and unordered, how about "Factors represent categorical data. They are stored as integers associated with labels and they can be ordered or unordered . While factors look (and often behave) like character vectors,

02-starting-with-data.Rmd
prosoitos: Indeed :stuck_out_tongue:

prosoitos:
Thanks for looking into this so quickly!

02-starting-with-data.Rmd
prosoitos: Oh, I see. I was trying to cut the repetition, but I cut it in the wrong place because I didn't know about the handout...

prosoitos:
For commit ac0796c, while some authors are clear in how they define "type" vs "class" (and this is what I followed in my commit comment and the reason of my suggested change), `typeof(x)` will return what these people would call "class". So I guess both are correct, depending of who you ask :stuck_out_tongue:

prosoitos:
Aaaaand I got my commit message wrong :stuck_out_tongue: 
It should read "Replace type with class", not the other way around. I did these quick changes while in the middle of my Carpentry instructor training, so I was being distracted :stuck_out_tongue:

fmichonneau:
Excellent work @prosoitos, thank you for your contributions!

prosoitos:
Thank you for looking into all of this so quickly @fmichonneau !! :slightly_smiling_face:

02-starting-with-data.Rmd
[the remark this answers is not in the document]
prosoitos: I like your additions on this section in your last commit: they split the novel concepts :) But I saw that you are keeping the c(xx:xx) :smile: Maybe I am wrong and it is an acceptable notation... :slightly_smiling_face:

PR #395, opened 2018-04-10
Make factor learning objectives more specific.

MilesMcBain:
During instructor training we discussed the learning objective 'manipulate categorical data' and found it quite vague. The factors section in this lesson is quite detailed and is clearly trying to get across multiple key points. I suggest changing learning objectives to include these points.


fmichonneau:
that looks good to me. Thanks!

PR #399, opened 2018-04-12
Minor edits to clarify terminology

shayofarrell:
Line 69: Change "variable" to "data frame" for consistent terminology. Also change the word "print" as it will be confusing to those new to R (or coding in general) as it has connotations of sending stuff to a printer.
Line 109: Change "Because columns are vectors, they all contain the same class of data", because it sounds like all columns in a data frame contain the same class of data.
Line 313: Change the US colloquialism "under the hood".
Line 316: De-italicize (is that a word?!) "levels" the second time it was done.
Line 484: Insert "function" in place of "call" which is a fluid representation that hasn't be used elsewhere.

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -66,7 +66,7 @@ surveys <- read.csv("data/portal_data_joined.csv")
 
 This statement doesn't produce any output because, as you might recall,
 assignments don't display anything. If we want to check that our data has been
-loaded, we can print the variable's value: `surveys`.
+loaded, we can see the contents of the data frame by typing its name: `surveys`.
 
 Wow... that was a lot of output. At least it means the data loaded
 properly. Let's check the top (the first 6 lines) of this data frame using the
@@ -106,7 +106,7 @@ spreadsheets from your hard drive (or the web).
 
 A data frame is the representation of data in the format of a table where the
 columns are vectors that all have the same length. Because columns are
-vectors, they all contain the same class of data (e.g., characters, integers,
+vectors, each column must contain a single type of data (e.g., characters, integers,
 factors). For example, here is a figure depicting a data frame comprising a
 numeric, a character, and a logical vector.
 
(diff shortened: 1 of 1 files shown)


02-starting-with-data.Rmd
```diff
@@ -481,7 +481,7 @@ surveys$plot_type <- factor(surveys$plot_type)
 > 2. Can you predict the class for each of the columns in the following example?
 >    Check your guesses using `str(country_climate)`:
 >      * Are they what you expected?  Why? Why not?
->      * What would have been different if we had added `stringsAsFactors = FALSE` to this call?
+>      * What would have been different if we had added `stringsAsFactors = FALSE` to the str() function?
```
ErinBecker: I think the point of this question is to ask what would have been different if `stringsAsFactors = FALSE` was added to the `read.csv()` call, not the `str()` call. I don't think `stringsAsFactors` is a valid input for `str()`.

02-starting-with-data.Rmd
```diff
@@ -66,7 +66,7 @@ surveys <- read.csv("data/portal_data_joined.csv")
 
 This statement doesn't produce any output because, as you might recall,
 assignments don't display anything. If we want to check that our data has been
-loaded, we can print the variable's value: `surveys`.
+loaded, we can see the contents of the data frame by typing its name: `surveys`.
```
ErinBecker: Has the term "data frame" been used at this point in the lesson? I think not, right? If not, maybe we shouldn't use it here yet.

02-starting-with-data.Rmd
shayofarrell: My bad! The call is question is data.frame() so changing the text so it refers to "data.frame()" as a function may be extra confusing given the lesson content.  Maybe leave as is or simply replace "call" with "function" without specifying the function name.?

02-starting-with-data.Rmd
shayofarrell: Yes, the section starts with a sentence containing "...we will use read.csv() to load into memory the content of the CSV file as an object of class data.frame". 
An alternative could be to add "object", as in "we can see the contents of the data frame object by typing its name: `surveys`."

shayofarrell:
Thanks François!

PR #402, opened 2018-04-18
"first rows" -> "last few rows"

stragu:
I am not sure if that error is due to a recent change in dplyr, but the rows are ordered so that the NA rows are at the bottom.
At least, this is what happens on my system, with the packages updated today.
Could someone confirm?


fmichonneau:
I see this too. Thanks for spotting it.

PR #403, opened 2018-04-18
add n variable like asked in challenge

stragu:
The challenge states: "Also add the number of observations (hint: see `?n`)"
I assume what is expected is to create an extra variable in `summarize()` that is simply equal to `n()`.


fmichonneau:
good catch, thanks!

PR #408, opened 2018-04-25
Added a #3 to the Challenge at the bottom of 01-intro-to-r.Rmd. Note …

John3491:
…that it could also be solved with sum() if na.rm=TRUE, but I did not add that answer because the concept of summing TRUEs was not taught


fmichonneau:
thanks @John3491 that's a good addition!

PR #415, opened 2018-05-02
clearer wording in "Indexing & Subsetting..." challenge

### LINKED ISSUE #414: Confusing wording in challenge question
@tavareshugo and I are running a workshop at the moment, and the wording of one of the challenges in the "Starting with Data" chapter has been confusing some of our learners.

The challenge in question is the first in the section "Indexing and subsetting data frames" ([here](https://github.com/datacarpentry/R-ecology-lesson/blob/master/02-starting-with-data.Rmd#challenge-1)). The wording of part 1 is unclear to our learners. 

> Create a data.frame (surveys_200) containing only the observations from row 200 of the surveys dataset.

Is the aim to take the 200th row of the dataframe? Or to take all rows from the 200th onwards?

My guess is that the intention was to pull out just the single row. Is that correct? Let me know, and, if you agree, I will be happy to submit a PR with a proposed new wording.

Thanks for all your work on these materials!

tobyhodges:
Fixes #414

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -278,7 +278,7 @@ In RStudio, you can use the autocompletion feature to get the full and correct n
 ### Challenges:
 ###
 ### 1. Create a `data.frame` (`surveys_200`) containing only the
-###    information in row 200 of the `surveys` dataset.
+###    data in row 200 of the `surveys` dataset.
 ###
 ### 2. Notice how `nrow()` gave you the number of rows in a `data.frame`?
 ###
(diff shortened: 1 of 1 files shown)


tobyhodges:
Perhaps "data" is a better word than "information"? I.e.

> 1. Create a `data.frame` (`surveys_200`) containing only the data in row 200 of the `surveys` dataset.

If you prefer this, I will push the change before you merge.

fmichonneau:
@tobyhodges yes, I prefer "data" here. Thanks!

tobyhodges:
Done

fmichonneau:
great thank you!

PR #417, opened 2018-05-04
Fix #409

### LINKED ISSUE #409: nrow(surveys) in R and SQL chapter
Hi,

Let me start with the fact that I think this is a wonderful and thorough lesson!

I think that there is a tiny code snippet that might be not what is intended. In the "R and SQL" chapter on the tab of querying databases with dplyr syntax, the last command is written as nrow(tbl), while I think you intended it to be nrow(surveys). In addition, the text below alludes to the "survey" rather than "surveys" data frame. Of course these are small corrections. Please let me know if I am misunderstanding something here.

Thanks.

-Elias

tavareshugo:
This should fix and close issue #409

### DIFF
--- a/05-r-and-databases.Rmd
+++ b/05-r-and-databases.Rmd
@@ -154,11 +154,11 @@ However, some functions don't work quite as expected. For instance, let's check
 how many rows there are in total using `nrow()`:
 
 ```{r nrows, results='show', purl=FALSE}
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks for the report @eoziolor and thanks for the fix @tavareshugo !

Issue #406, opened 2018-04-24
Starting with data: Remove use of class()


humburg:
There are a few instances in the first part of *Starting with data* where the class of a data frame is examined. Sometimes directly, using `class()` and sometimes indirectly using `str()`. The fact that the class of a data frame is `data.frame` isn't really relevant to the objective of the lesson (learning how to interact with data frames). I would like to propose removing all uses of `class()` as well as the question about the class of a data frame. This would help avoid possible confusion about a detail that isn't relevant to the lesson and keep the focus new contents learners need to digest.

I'd be happy to put together a pull request to this effect if there is agreement that the change should be made.

mondorescue:
@humburg Thanks for sharing your thoughts. From what I see on the _Starting with data_ page, `str()` is being used for examining the data frame. For clarification, are you proposing the removal of the _Inspecting `data.frame` Objects_ section?

humburg:
@mondorescue I'm not proposing to remove the entire *Inspecting `data.frame` Objects* section. Most of that content is relevant, I think. I'm only suggesting to remove the references to `class()`. For example:

> * Object class:
    * `class(surveys)` -  information about the class of the object

There are a few more places where the class of a data frame is mentioned. All in all it is a pretty small change.

Anyone who understands what `class()` does will not find it surprising that a `data.frame` is a `data.frame` and won't gain any deeper insight into how data is stored in data frames or how these are manipulated. But someone who doesn't quite understand the concept of classes might be confused by how this relates to the actual objective of the lesson. I think it is an unnecessary distraction.

aurielfournier:
I see @humburg point as well, and when thinking back to when I've taught that part of the lessons I know I always paused to talk about how this is one of the many ways that we need to check our inputs/outputs along the way to make sure R and the user are on the same page, and I think its very valuable in that case, but `str()` might be just as useful for such a reminder.

PR #419, opened 2018-05-06
Remove reference to `class()`

humburg:
As discussed in #406, using `class()` in *Starting with data* is not strictly necessary and could be a source of avoidable confusion.


fmichonneau:
this looks good. Thanks!

Issue #421, opened 2018-05-09
Suggestion to use mapping=aes(...) in ggplot section.


pfh:
(This issue submitted as part of my instructor training checkout.)

Introducing ggplot syntax, the template given is:

`ggplot(data = <DATA>, mapping = aes(<MAPPINGS>)) +  <GEOM_FUNCTION>()`

(Note: this should be enclosed in ` ``` ``` ` in order to render to html correctly.)

However when R code is given, the `mapping = ` is dropped:

`ggplot(data = surveys_complete, aes(x = weight, y = hindfoot_length))`

This will work, but it is very unusual to give a positional argument after a named argument (Python forbids this entirely, but it seems to work in R). This seems likely to cause confusion, as the `aes` function is already somewhat mysterious.

aurielfournier:
Hello @pfh thanks for opening this issue.

Coming from python, a language I'm not fluent in by a long shot, that may seem odd, but is quite common in R, and we are using the mapping= to help guide the verbal description of the code, when explaining what `aes()` actually does.

Quite frequently the mapping= argument label is not used when folks are actually writing ggplot code, and so we want the learners to be comfortable seeing it as a positional argument, since many of the examples they will find online are written this way. 

If you think we could improve the explanation of this in our written lessons I would encourage you to submit a pull request.

PR #422, opened 2018-05-09
Use mapping=aes(...) in ggplot section.

pfh:
This is my suggested fix for my suggested issue #421

mapping= is added to all uses of ggplot where data= is used.

ggplot template is surrounded by ` ``` ``` `.

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -58,7 +58,9 @@ this fashion allows for extensive flexibility and customization of plots.
 
 To build a ggplot, we will use the following basic template that can be used for different types of plots:
 
+```
 ggplot(data = <DATA>, mapping = aes(<MAPPINGS>)) +  <GEOM_FUNCTION>()
+```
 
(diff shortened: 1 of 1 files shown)


aurielfournier:
#421 tagging in my response to @pfh on the issue he opened about this. 

I'm not sure this is the right way to go, as the mapping= argument is not required, and many examples [including the tidyverse tutorials built by the people who built ggplot, http://ggplot2.tidyverse.org/] don't include the named argument. 

What do other maintainers think?

mondorescue:
To me @pfh's PR addresses the inconsistency in the lesson on how arguments in the `ggplot()` are assigned. We should either have the examples exclude the `data =` argument (to be consistent with implicit `mapping =` argument) or include the `mapping =` argument. I lean towards including the `mapping =` argument. In "R for Data Science", Hadley explicitly assigns the arguments (e.g. http://r4ds.had.co.nz/data-visualisation.html)

aurielfournier:
alright, if its used there, I can deal with it. I've just never seen it used before, but that is just my experience.

PR #429, opened 2018-05-22
Count for combination of factor in the dplyr lesson

gederajeg:
I add two code-chunks in the "03-dplyr.Rmd" lesson in the **counting** section.
The first chunk shows how to generate **co-occurrence** count between _a set of factors_ (i.e. `sex` and `species`) using `dplyr::count()`, complementing the example of `dplyr::count()` for a single factor (i.e. `sex`). 
The second chunk shows the use of `dplyr::arrange()` to sort the co-occurrence count given _a set of criteria_, complementing the previous example for `dplyr::arrange()` before the **counting** section.

### DIFF
--- a/03-dplyr.Rmd
+++ b/03-dplyr.Rmd
@@ -409,6 +409,23 @@ surveys %>%
     count(sex, sort = TRUE) 
 ```
 
+Previous example shows the use of `count()` to count the number of rows/observations for *one* factor (i.e., `sex`). If we wanted to count *combination of factors*, such as `sex` and `species`, we would specify the first and the second factor as the arguments of `count()`:
+
+```{r purl = FALSE}
+surveys %>%
+  count(sex, species) 
+```
+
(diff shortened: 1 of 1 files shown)


fmichonneau:
This looks good to me. Could another maintainer also review and merge?

aurielfournier:
Looks good to me as well. Thanks @gederajeg !

orchid00:
Hi, I'm not a maintainer, but I'm going to teach the lesson, so I want to give my 2 cents.

1)Thinking about cognitive load, would rather not add more.

2) In case it is added, I suggest to add an optional tag - especially for new instructors who follow the notes line by line. 
Also please don't introduce in the exercise the use of n() as it hasn't been taught before. 
Slight grammatical corrections will be needed in the text too.

aurielfournier:
Thanks for the input @orchid00 , would you be up for submitting a PR with those grammatical corrections?

PR #435, opened 2018-06-12
fixed typo in unique genera code

MCMaurer:
Just a small typo when creating `species` from the database, it threw an error when a different name was used later on.


mondorescue:
Thanks for catching this @MCMaurer .

PR #439, opened 2018-06-19

saramannheimer:
I added more detail about the download.file command. This suggestion is part of my instructor checkout. Let me know if you have questions!


anacost:
Hi @saramannheimer, thank you!

PR #441, opened 2018-06-24
Have added a section for the group object. 

Math2522:
This section introduces the grouped object which is created with the group_by() function. It shows that summarize preformed on a grouped object is different as summarize preformed on an ordinary dataframe. It continues to show the use of the ungroup() function. Further, added a comment about overriding existing grouping in the end of the summarise section.

### DIFF
--- a/03-dplyr.Rmd
+++ b/03-dplyr.Rmd
@@ -303,6 +303,36 @@ then combine the results. **`dplyr`** makes this very easy through the use of th
 `group_by()` function.
 
 
+#### The `group_by()` object
+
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks for your contribution @Math2522!
The presentation of `group_by()` that you introduce here is useful but it's a little disconnected from the following section about the `summarize()` function. Additionally, we are limited in the amount of material we can add to this lesson which already has quite a bit of content. If you'd be interested in reworking your presentation of `group_by()` in combination with `summarize()` please let me know, I'd be happy to work with you on it.

PR #458, opened 2018-08-01
updated writing databases section

callinSwitzer:
1. downloaded the species, plots, and surveys data into the data folder. 
2. imported tidyverse, to be able to use the read_csv() function
3. changed output file to "data_output/portal-database-output.sqlite" to be consistent with other lessons
4. added "copy_to(my_db, species)" because it was missing


callinSwitzer:
I just realized "copy_to(my_db, species)" would be part of the final challenge.

PR #460, opened 2018-08-16
minor additions in R-ecology ggplot lesson

pratee11:
- This contribution is a part of the instructor training checkout process.

There are two changes (additions) made in the file:
1. The lesson specified arguments for ggplot as data and mapping every time. I mentioned that data = and mapping = are not needed until they are in correct order, which they usually are. It will help make the codes shorter and cleaner.
2. I introduced the "up" key to display the previous command on R console. It will save a lot of typing, especially when we are making minor changes and going back and forth on a plot. It will save time in the workshops, and also prevent typos.

---

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -118,6 +118,10 @@ surveys_plot +
 - Anything you put in the `ggplot()` function can be seen by any geom layers
   that you add (i.e., these are universal plot settings). This includes the x- and
   y-axis mapping you set up in `aes()`.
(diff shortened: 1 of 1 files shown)


sabahzero:
Include another Challenge after "Intro to R > Functions and their arguments", though it may be too complex as an introductory exercise?

Q: What script mimics the same function as sqrt()?
A: x^0.5, where x is any positive number
Q: How might you write this as a function such that the output is the same as sqrt()?
A:
```
squareroot = function(x) {
x^0.5
}
squareroot(x) = sqrt(x)
```

mondorescue:
@pratee11 thank you for your suggestions -- we will take it into consideration as we make changes to the lesson.

PR #468, opened 2018-09-14
Small typo fixes and add an intermediate object

kbjornen:
Adding an intermediate object allows the mutation to be checked. OW it is not visible because it goes off the desktop.


mondorescue:
@kbjornen thank you for your contribution.

```
surveys %>%
  mutate(weight_kg = weight / 1000)
```
itself outputs the result to the console, so it isn't necessary to have an intermediate object.

Your comment on `%in%` will be addressed in our **tidyverse_first** branch.

PR #470, opened 2018-09-18
added indexing.svg (and indexing.pdf, just in case) as a toy example …

mkamenet3:
…visualization for how indexing works in a data frame. I also added a blurb under "Indexing and subsetting data frames" explaining the toy example. During the lesson, I connected this toy  example back to the SQL lesson on how we want data to be organized in a rectangle. *NOTE:* I couldn't get the image to fully render when I knit the .Rmd file, however could not fully figure out why (the space for the image renders in the HTML, but the actual image does not). I thought it could be an issue with my computer. That's why I've also uploaded *indexing.pdf* just in case. Please let me know if there are issues with the image on your end. Thank you! 

PR part of instructor training checkout.


fmichonneau:
Thank you for your contribution @mkamenet3. While I find that analogies can be useful to help learners make connections in their mental models when presenting new concepts, as it is currently presented, the drawing adds additional complexity that would make the introduction of indexing for data frame maybe more confusing. I'm going to close this pull request but if you are interested in revisiting this, please let me know and I'd be happy to work with you on an alternate analogy.

PR #513, opened 2019-01-04
add facet continuous data challenge

Irosenthal:
Added a brief challenge exercise to show how to use cut() to facet ggplots by splitting continuous data.


fmichonneau:
Thanks for your contribution @Irosenthal. While the `cut()` function is useful, I think it's out of scope for this lesson aimed at beginners. We already introduce a lot of content in this ggplot2 lesson.

PR #519, opened 2019-01-25
Incorrect assignment of new column

mand35:
the original statement `survey%date <- ...` lead to `Error: unexpected input in "surveys%date <-`,


mand35:
Sorry, my fault, I mixed up % and $.
Everything works fine already with the original version.

PR #526, opened 2019-03-04
Url updates

njahn82:
Some updates for URLs that have recently changed. Don't panic, the old URLs still work and, as in the case of links to the ggplot2 docs, are re-directed!


fmichonneau:
thank you!

PR #534, opened 2019-04-15
Update 02-starting-with-data.Rmd for csv attributes

martinjeanphd:
Add text in the note section about csv file to invite users to pay attention of other attributes of csv files:

- character encoding (important for international users);
- control characters used for line ending;
- date format (if note splitter into three variables)
- presence of unexpected newlines


fmichonneau:
thanks for this addition

Issue #536, opened 2019-04-29
Lesson suggestions (submitted by email from trainee)


ErinBecker:
The following are my contributions in R lesson, as the last check out step to become an instructor.
First the lesson is a bit difficult for a beginner to follow. Example the session "before we start" should be part of introduction session by summaries into a short sentences,the way its in the "introduction to R" for a beginner to follow.
Second: In the session "manipulating data frames", before: library("tidyverse"), I suggest to add, install.packages("tidyverse"). This step have been mentioned in the previous lesson by provided a link to follow, but not all will able to follow or to get that, and then short sentence which says do this step or install this package if you have not done so in the previous lesson.

dklinges9:
I had the same thought when I recently went through the lesson. The only point in the lesson that the command `install.packages()` is mentioned is briefly at the bottom of the "Home" index....and it isn't quite mentioned with full context (the page says to type the command into the console, but what the "console" is hasn't actually been introduced yet.
In an actual in-person workshop, using `install.packages()`would certainly come up at least once, but for learners playing from home it might be nice to mention again.

I [submitted a pull request](https://github.com/datacarpentry/R-ecology-lesson/pull/538) in response to the above comment and my thoughts presented here.
Note that I didn't include a code chunk with `install.packages()`, as this would have required setting a CRAN mirror and looking through the repo it seems as if this is something intentionally avoided...

fmichonneau:
fixed by #538

PR #538, opened 2019-05-08
Install packages fixes

dklinges9:
A few quick suggested fixes, partly in response to #536 and also from some personal notes.


fmichonneau:
I think having a reminder about this here makes sense. Thanks for the PR!

PR #539, opened 2019-05-13
updating, instructor training checkout

melimore86:
updating based on Data Carpentry instructor training checkout, Melissa Moreno, melimore86@ufl.edu, https://carpentries.github.io/instructor-training/20-checkout/index.html


fmichonneau:

PR #540, opened 2019-05-14
Prove why order/naming arguments in functions matters

ValentineHerr:
To me the real proof that you need to name the arguments if you enter them in a different order is to show what happens when you don't name them and put them in a different order.


anacost:
Hi @ValentineHerr, 
thank you for your pull request! This is a good comment to be made during the workshop if someone asks about how to work with functions and their arguments. 
The lesson is already very rich in details. This topic is not essential. Therefore, I will close this pull request.

PR #548, opened 2019-06-25
typo: removed extra period at end of sentence

dklinges9:
Simple fix of removing an extra period in .Rmd file.


mondorescue:
@dklinges9 Thanks for catching this.

PR #549, opened 2019-06-25
Clarify logic examples

dklinges9:
This edit is in response to confusion by learners at a workshop I co-instructed last week. In the previous version, the second example of combining boolean operators:
```
weight_g[weight_g >= 30 & weight_g == 21]
```
returned `#> numeric(0)` because a value cannot be >= 30 AND == 21. Because this is the first exposure to conditionals for most learners, an affirmative example, that doesn't return a NULL value, would be better. I've changed both examples so that all the same boolean operators (`>`, `<`, `<=`, `==`) are still used, and both examples return valid vectors.

### DIFF
--- a/01-intro-to-r.Rmd
+++ b/01-intro-to-r.Rmd
@@ -467,6 +467,7 @@ You can combine multiple tests using `&` (both conditions are true, AND) or `|`
 ```{r, results = 'show', purl = FALSE}
 weight_g[weight_g > 30 & weight_g < 50]
 weight_g[weight_g <= 30 | weight_g == 55]
+weight_g[weight_g >= 30 & weight_g == 21]
 ```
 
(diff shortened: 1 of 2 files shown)


mondorescue:
@dklinges9 I think the intention was to inform students what would happen if the conditionals conflicted (i.e. it would return a `numeric(0)`).

@datacarpentry/r-ecology-lesson-maintainers What do you guys think? Should we change the example so that it doesn't return a NULL and maybe add a line in the instructor's note to do a demo of a NULL output?

fmichonneau:
I think we can add an example of a call that returns something, but keep the example that returns an empty vector. I think it's useful to show this even to beginners so they'll know what the problem might be.

fmichonneau:
Thank you for your contribution @dklinges9!

PR #562, opened 2019-09-05
Improve first mutate example

katrinleinweber:
Multiplying by 2 seems to be a less realistic example than the [one we have in 01-intro-to-r.Rmd](https://github.com/datacarpentry/R-ecology-lesson/blob/ed1c4aeac050cb49f7c01f9a2889a7e9555b7309/01-intro-to-r.Rmd#L123). Should we re-use that one?


fmichonneau:
yes, absolutely. Thank you!

PR #563, opened 2019-09-06
change data and raw data paths

katrinleinweber:
Would it make sense to piggy-back laziness by making the name of the raw data folder, in which people shouldn't meddle with files, longer?

This PRs is a quick search-and-replace attempt to do that, so that `data` is now for the output.

I also considered the idea of introducing a `data/raw` sub-folder, but since the output data's folder should be completely deletable, I guess `data_raw` on the same level is better.

What do you think? Implementing this would probably require

- [ ] sending a coordinated set of PRs against [other DataCarpentry repos](https://github.com/search?o=desc&q=org%3Adatacarpentry+data_output&s=indexed&type=Code).


fmichonneau:
that sounds like a reasonable change to me. What do the other @datacarpentry/r-ecology-lesson-maintainers think?

mondorescue:
I like it. It will make it easier to wipe the outputs, if needed.

katrinleinweber:
I will need a few more days to test the lesson. I might force-push some more instances of paths that need to be updated here. Will remove the `WIP:` from the title then. Please feel free to set it into draf-PR mode if that's possible.

anacost:
I like it too.

katrinleinweber:
OK, I'm pretty sure I got them all now.

katrinleinweber:
The handout had a few more instances, but `source("make_code_handout.R")` created a huge diff. I updated the file manuall now.

After this PR, should we do separate one with only the output of `source("make_code_handout.R")`? Or an automation, that let's Travis create the handout?

fmichonneau:
the code handout on the website is generated by Travis. You probably saw a big diff in the one in the repo because it hasn't been updated in a while. I'm wondering whether we should even remove it from the master branch... In any case, thanks for this!

PR #564, opened 2019-09-10
Replace x/ylab() with labs(x/y = ...) to reduce variation / cognitive load

katrinleinweber:
The former are nice shortcuts, but repetition of a single (and their) "parent" function is probably better for novices.

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -598,14 +598,14 @@ library(gridExtra)
 spp_weight_boxplot <- ggplot(data = surveys_complete, 
                              mapping = aes(x = genus, y = weight)) +
   geom_boxplot() +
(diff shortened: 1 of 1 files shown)


04-visualization-ggplot2.Rmd
```diff
@@ -598,14 +598,14 @@ library(gridExtra)
 spp_weight_boxplot <- ggplot(data = surveys_complete, 
                              mapping = aes(x = genus, y = weight)) +
   geom_boxplot() +
-  xlab("Genus") + ylab("Weight (g)") +
   scale_y_log10() +
+  labs(x = "Genus", y = "Weight (g)") +
```
katrinleinweber: I moved this downwards, because all other `labs()` are directly above a `theme` function.

Issue #545, opened 2019-06-15
update ggplot2 facet syntax 


tavareshugo:
The newer versions of `ggplot2` (`>3.0.0`) now support the following syntax for facets:

```r
# old
myplot + facet_wrap(~ species_id)

# new
myplot + facet_wrap(vars(species_id))
```
and

```r 
# old
myplot + facet_grid(sex ~ .)

# new
myplot + facet_grid(rows = vars(sex))
```

This new syntax seems more intuitive, and I often found that learners get a bit thrown off by the `~` notation, which seems too exotic given they didn't learn about formula notations. 

Before making a pull request, I wonder if others have any other pros/cons about updating the materials with this new notation instead? (apart from the fact one has to ensure people have updated their `ggplot2`)

dravery:
Totally agree. Perhaps it would be useful to include both options or a short description of the formula notation mapped to the new vars() syntax.

monicagerber:
I also agree. I taught this last week and it was confusing to introduce the formula notation at the end of the lesson. Updating to the new syntax will also more closely follow what is in the `ggplot2` [reference](https://ggplot2.tidyverse.org/reference/facet_grid.html) for `facet_grid()`. 

Here is my suggestion for the lesson.

Old:
```
The `facet_wrap` geometry extracts plots into an arbitrary number of dimensions
to allow them to cleanly fit on one page. On the other hand, the `facet_grid`
geometry allows you to explicitly specify how you want your plots to be
arranged via formula notation (`rows ~ columns`; a `.` can be used as
a placeholder that indicates only one row or column).
```

New:
```
The `facet_wrap()` geometry arranges a one-dimensional sequence of panels to allow 
them to cleanly fit on one page. On the other hand, the `facet_grid()` geometry allows 
you to form a matrix of rows and columns of panels. Both geometries allow to to specify
 faceting variables quoted with `vars()`. For example, `facet_wrap(vars(facet_variable))` or 
`facet_grid(vars(row_variable), vars(col_variable))`. The historical interface using formulas
 is also available  (`rows ~ columns`; a `.` can be used as a placeholder that indicates only 
one row or column).

mondorescue:
@tavareshugo would you be willing to make a PR with @monicagerber 's text suggestion?

tavareshugo:
@mondorescue I'm happy to do a pull request (working on it now)

[...]

PR #565, opened 2019-09-10
Reduce facet_...() compatibility note to NEWS reference

katrinleinweber:
Related to #545 & #551. ggplot2's faceting update is 1 year old now. I think mentioning both syntaxes was useful for a time, but going forward, the old one might be more confusing than useful to our learners.

I'm not sure whether that point is now or maybe another year in the future. Please feel free to delay this PR if a longer transition time seems more appropriate.


fmichonneau:
that's useful, thank you!

katrinleinweber:
You're welcome & thanks for merging :-)

PR #567, opened 2019-09-10
Remove duplicate directory image

katrinleinweber:
Follow-up to #563. `img/r-starting-how-it-should-look-like.png` has the workshop-relevant structure & is more maintainable, because it's a screenshot.


fmichonneau:
thanks!

PR #573, opened 2019-09-12
Align handout comments with challenges

katrinleinweber:
The `r` blocks with the challenge questions texts are sometimes out-of-sync with the `{.challenge}` boxes. Here's one example. Since I'm on vacation for two weeks, anyone who wants to continue this, is welcome to it :-)

Follow-ups to: 041b9a97

### DIFF
--- a/01-intro-to-r.Rmd
+++ b/01-intro-to-r.Rmd
@@ -364,11 +364,9 @@ factors (`factor`) and arrays (`array`).
 > ```
 >
 > * How many values in `combined_logical` are `"TRUE"` (as a character) in the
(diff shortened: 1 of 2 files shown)


katrinleinweber:
For the long-term: Could we maybe adjust [`make_code_handbout.R`](https://github.com/datacarpentry/R-ecology-lesson/blob/33f8917a177674e6cbdddcb2770935bf49a350bd/make_code_handout.R) to reg-ex-extract every line below a `{.challenge}` that starts with `> ` but not with `> ```{r`?

PR #576, opened 2019-09-30

warutherford:
The introduction to the data set (line 33) for this lesson was a little confusing as an ecologist. "Species repartition" is not a commonly used phrase when working with animal data, which may have an ecologist confused from the start. I clarified the language to include more common phrasing that an ecologist new to R may understand more clearly.

Line 241: c() argument not needed and may be confusing when not part of the "Indexing and subsetting data frames" challenge solution with nrow().

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -30,8 +30,8 @@ source("setup.R")
 ### Presentation of the survey data
 ```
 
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thank you so much for this contribution. The updates to the language make the presentation of the data clearer!

PR #579, opened 2019-10-02
Trim "Asking for help" to the essentials

katrinleinweber:
Similar to https://github.com/swcarpentry/python-novice-gapminder/pull/443, but since the resources we mention cover much of the advanced content here, how about we reduce this section to the most relevant part for new learners?


mondorescue:
I like the idea of streamlining the "Asking for help" section. @katrinleinweber I think you did a great job here. I support merging this PR.

@datacarpentry/r-ecology-lesson-maintainers What do you guys think?

katrinleinweber:
Thanks @mondorescue :-) I noticed a typo just now and re-wrote a small part around it.

PR #584, opened 2019-11-15
[fix #583] explain how to combine pipe operator for ggplot

### LINKED ISSUE #583: Add piping operator in ggplot lesson
I just gave a *Data Wrangling in `R`* using *Data Analysis and Visualization in R for Ecologists* and am currently preparing to give the *Data Visualization* follow-up workshop. While preparing, I noticed that nowhere in the Data Visualization materials do we use the pipe operator (`%>%`) to pipe a dataset into `ggplot()`. 

I think adding a short blurb on how you can do this would make a great connection between the Data Wrangling and Data Visualization workshops!

SaraTomiolo:
I have tried to address the issue and added a small explanation and a couple of examples of how the pipe operator can be integrated with ggplot2 obtaining the same results and in the previous plot (where no pipe is used)

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -355,16 +355,17 @@ ggplot(data = yearly_counts, mapping = aes(x = year, y = n, color = genus)) +
 
 ## Integrating the pipe operator with ggplot2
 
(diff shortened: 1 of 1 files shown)


fmichonneau:
thanks for this! I updated the language a little but it looks good.

PR #585, opened 2019-12-05
fixed typo in ggplot lesson

lisamr:
Within the ggplot plot facet lesson, there is typo in the comments to group the data by species_id. It should be by genus.


mondorescue:
@lisamr thanks for catching this.

PR #595, opened 2020-01-19
Clarification in the split-apply-combine section

FloHu:
Dear Carpentries Team,

When going through the split-apply-combine section I found the paragraph where grouping by multiple columns is introduced slightly confusing. First, how to look at the last rows of a data set is not introduced before the section. Second, when talking about "last column" it is not clear if it refers to the original surveys data or the summarized version. And finally I felt like some clarification of why the call to mean results in NaN might be helpful. 

Thanks and best wishes,
Florian


fmichonneau:
Thanks Florian! Your edit makes this section much clearer.

Issue #450, opened 2018-07-09
replace gridExtra with patchwork


fmichonneau:
Once [`patchwork`](https://github.com/thomasp85/patchwork) gets released on CRAN, use it to combine multiple ggplot plots instead of relying on `gridExtra`.

camillanovaglio:
Hi, 

I'd like to work on this as my first issue to checkout. Is this about replacing gridExtra with patchwork in 04-visualisation-ggplot2.Rmd line 591 and following example? 

Thanks

fmichonneau:
@camillanovaglio, @johnsonra recently contributed the fix for this. If you have other ideas on how to build upon this contribution or improving the ggplot2 lesson in general, we'd love to review your contribution! Thank you!

camillanovaglio:
Great, thanks! I have just sent a contribution to another topic.

PR #596, opened 2020-02-10
Replace gridExtra with patchwork

johnsonra:
Re issue #450: Replaced use of gridExtra with patchwork for combining ggplot graphics.


fmichonneau:
this looks great, thank you for your contribution!

PR #597, opened 2020-02-13
Spacing, function consistency, and language clarification

atheobold:
In these commits: 

1. The spacing was changed to have consistent spaces before and after = signs.
2. The line widths were returned after 80 characters.
3. The `read_csv( )` function was added to be consistent with the `tidyverse` tools used. 
4. Additional language was added surrounding aesthetics, 
5. An expression was added to an axis label with a log transformation.


rmmart17:
I appreciate the suggestion to create a new line after 80 characters. It is easier to read, especially on the mobile device that I have. 
Your edits also have minimal "jargon", and focuses on the code. Overall, I think your edits are an improvement to the lesson.

fmichonneau:
thank you

PR #618, opened 2020-06-18
updating to main

JKRWard:
updating instructions to reflect the commitment to using 'main' rather than 'master' for branch


fmichonneau:
thanks!

PR #619, opened 2020-06-21
Small fixes to ggplot2 episode

csoneson:
* Remove out-of-context comment about `theme_bw()` in the ggplot2 episode.
* Replace `group_by()` + `tally()` by `count()` in the ggplot2 episode since `tally()` was not previously introduced or used. From what I can see, the grouping in `yearly_sex_counts` is not required, so it felt easier to not introduce a new function.


fmichonneau:

PR #620, opened 2020-06-23
Urgent: link fix (old link redirects to incorrect location)

k8hertweck:


fmichonneau:
thanks Kate, I hadn't seen your PR before fixing it.

PR #622, opened 2020-06-29
include script file saving - typo fix attempt 1

bridgeta:
...


bridgeta:
Still figuring out Github
this was in error

fmichonneau:
@bridgeta thanks for incorporating the suggested changes. Before I merge this, there still the issue of the shortcut for saving. The text says "Shift+S" but I think it's "Ctrl+S" (and "⌘ + S" for macOS). Which operating system do you use and do you use "Shift"?

bridgeta:
I see now. I'll make the change.

fmichonneau:
replaced by #625

PR #634, opened 2020-08-30
shorter pronounciation of <-

steltenpower:
Also makes saying things in the order they're written.


fmichonneau:
I have never heard of this but I find this useful!
Could another @datacarpentry/r-ecology-lesson-maintainers confirm they are OK with this change?

mondorescue:
I like this. I second the change.

PR #635, opened 2020-08-30
helping to remember + talk about c() with "chain"?

steltenpower:
As concatenate is rather long ... and to many people means adding strings together in a way that can't be undone

### DIFF
--- a/01-intro-to-r.Rmd
+++ b/01-intro-to-r.Rmd
@@ -272,7 +272,7 @@ doing.
 A vector is the most common and basic data type in R, and is pretty much
 the workhorse of R. A vector is composed by a series of values, which can be
 either numbers or characters. We can assign a series of values to a vector using
-the `c()` function. For example we can create a vector of animal weights and assign
+the `c()` function (which some remember or pronounce as "Chain"). For example we can create a vector of animal weights and assign
 it to a new object `weight_g`:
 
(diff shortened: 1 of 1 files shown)


steltenpower:
Have a better idea?

fmichonneau:
could any other @datacarpentry/r-ecology-lesson-maintainers confirm they are OK with this change?

mondorescue:
@fmichonneau I would stay close to the documentation of `c()`, so if "concatenate" is too long, then use "combine" instead. "Chain" is something new. I would reject this PR.

steltenpower:
I'm all for staying with terminology used elsewhere and all for consistency within a carpentry lesson. But my limited me hasn't seen the one super-clear and consistent thing to stick to.
Why I chose "chain": you can still see the individual parts that went into it; a metal chain consisting of links. Without keeping extra administration you can reverse the operation, which is not the case for a string concatenation.

Teebusch:
Concatenate means ["to link together in a series or chain"](https://www.merriam-webster.com/dictionary/concatenate), so semantically it isn't very different from "chain".

Anyways, the documentation for `c()` says "This is a generic function which **combines** its arguments." so I would agree with @mondorescue - stick with the official docs and call it "combine".

mondorescue:
We will close this PR.  @steltenpower thank you for the suggestion.

PR #637, opened 2020-08-31
explain how genus and genera relate

steltenpower:
This lesson is very useful too for people not ecologists.
With little subtle changes this can be improved, this is an example.


fmichonneau:
that's a useful clarification. Thank you.

PR #638, opened 2020-08-31
turned referal to SQL lesson into link

steltenpower:
Especially useful for those not going through the whole curriculum and in the same order


fmichonneau:
thank you for this useful addition!

PR #639, opened 2020-09-01
Update 00-before-we-start.Rmd

zacchaeus-compson:
I added one objective to the "Learning Objectives" to better match the initial content of the lesson.

Note that this edit was made, in part, for my Data Carpentry Instructor Training checkout.


fmichonneau:
Thank you!

PR #647, opened 2020-10-02
remove quotation marks in `library()` calls

Teebusch:
Quoting package names in `library()` is not necessary and disables auto-completion of package names in RStudio.


mondorescue:
@Teebusch I agree with you. I was planning to do it... Thanks for editing this.

PR #648, opened 2020-10-02
closes #586 confusing example

### LINKED ISSUE #586: 02-starting-with-data.Rmd line 226 example confusing
code below is confusing. One of the first 3 elements in the 7th column has a blank, so only 2 elements are returns. It's extra confusing because we haven't spoken about factors yet. Switching it to the surveys[1:3, 6] will make it clearer. 
```
# first three elements in the 7th column (as a vector)
surveys[1:3, 7] 
```

Teebusch:
when the 7th column is a factor vector, the result has two elements, although 3 were requested, because one element is blank. With the 6th column this is not the case.
Also changed "elements" to "rows" for consistency.

closes #586

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -223,8 +223,8 @@ surveys[1, 6]
 surveys[, 1]    
 # first column of the data frame (as a data.frame)
 surveys[1]      
-# first three elements in the 7th column (as a vector)
-surveys[1:3, 7] 
+# first three rows of the 6th column (as a vector)
+surveys[1:3, 6] 
 # the 3rd row of the data frame (as a data.frame)
 surveys[3, ]    
 # equivalent to head_surveys <- head(surveys)


mondorescue:
Will 7th column still be a factor vector with R v4.0, since in version 4, stringsAsFactor defaults to FALSE? I haven't installed R v4.0 to check myself.

Teebusch:
No, it wont. In R >4.0 none of the character columns will be converted to factors by default. In that case, both the old and new version (i.e., 6th and 7th column) will return a length 3 character vector. Accessing the 6th column works as intended with old and new versions of R and is therefore prefereable. Whether the returned vector is a character vector or factor vector is irrelevant for the learning objective, but returning 2 elements when 3 are requested (current lesson with R <4.0) is unnecessarily confusing.

The change in default `stringsAsFactors` is a bigger issue that should be addressed, too, but separately.

mondorescue:
@Teebusch Sounds good to me. Will approve PR.

PR #649, opened 2020-10-02
closes #471 unclear question "How many species...?"

### LINKED ISSUE #471: Potentially confusing question in Inspecting Data Frames Challenge
A minor issue came up while I was teaching the "Starting With Data" lesson. After inspecting the data frame, there is a Challenge which asks, among other things: "How many species have been recorded during these surveys?"

This is straightforward kind of question to answer, but what is a bit confusing is that there are both "species" and "species_id" columns in the data set. If you think about it a bit, it's clear the answer is that you should use "species_id" column, because some species names are repeated in multiple genera. But thinking about this and discussing it with the class seemed to me to add cognitive load that was not helpful to meeting the goal of the lesson. 

I suggest substituting the question:

Teebusch:
The species column has fewer unique entries than species_id. The relation between genera and species can be unclear to learners without biology background. Asking about genera fulfills the same learning objective but is less ambiguous.

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -174,7 +174,7 @@ objects besides `data.frame`.
 >
 > * What is the class of the object `surveys`?
 > * How many rows and how many columns are in this object?
-> * How many species have been recorded during these surveys?
+> * How many genera have been recorded during these surveys? (Hint: genera is the plural of genus).
 >
 > ```{r, answer=TRUE, results="markup", purl=FALSE}
 >
@@ -182,7 +182,7 @@ objects besides `data.frame`.
 >
 > ## * class: data frame
 > ## * how many rows: 34786,  how many columns: 13
-> ## * how many species: 48
+> ## * how many genera: There are 26 unique genera in the `genus` column."
 >
 > ```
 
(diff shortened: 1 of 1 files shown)


mondorescue:
@Teebusch FYI: I think your suggestions are sound and I would approve merging it, but to change the objectives, it requires the CAC to review it -- if I recall correctly.

@fmichonneau Am I correct?

Teebusch:
@mondorescue I see. I'd argue that the learning objective doesn't change (inspecting the output of summarize to see how many levels a factor has). It's simply asking for a different factor (`genus` instead of `species`), because there's some ambiguity regarding `species` and `species_id`.
This has tripped up multiple instructors (including me, as I went through the lesson), so I'd hope it can be resolved quickly.

Teebusch:
I think this PR can be rejected/closed, since the challenge has changed with #663 
The challenge has moved to a later point in the lesson, after the concept of factors has been introduced, and is now asking for the number of genera
The challenge is now:

> 1 Change the columns taxa and genus in the surveys data frame into a factor.
> 2 Using the functions you learned before, can you find out...
>  * How many rabbits were observed?
>  * How many different genera are in the `genus` column?

Learners are expected to use the `summary` and `nlevels` functions to answer the questions

```
surveys$genus <- factor(surveys$genus)
summary(surveys)
nlevels(surveys$genus)
```

(note: there's a little markdown formatting issue in the challenge that needs to be fixed)

fmichonneau:
thanks!

PR #650, opened 2020-10-02
closes #628 axis labes not showing when using plot on Mac

### LINKED ISSUE #628: Resolution to issue of plot(as.factor(surveys$sex)) not showing axes
With teaching a recent workshop, we came across an issue where several individuals with old Macs could not get their axes labels to show up when calling the plot() function (no error was outputted, the plot axes just never showed up).  Our workshop coordinator mentioned that they had previously encountered this issue and believe that it is due to R not being able to find the default Arial font. 

To resolve this issue of no axes text: Go to Finder Search for "Font Book" and open it. Look for the Arial font and, if it is greyed out, turn it on. This eliminates the warning and enables R to produce the appropriate labels.

Alternatively, we had success with this issue when plotting with ggplot2 by specifying a specific font type using theme. I'm not sure if there is an equivalent to plot(), but if there is, that could be a shorter workaround. 

Perhaps we could add a note to the Instructor Notes for how to resolve this if it comes up?

Teebusch:
Apparently on some MacOS versions Arial is deactivated. This can cause issues with drawing axis labels in R. I've added instructions how to activate Arial to the instructor notes.

### DIFF
--- a/instructor-notes.md
+++ b/instructor-notes.md
@@ -61,6 +61,13 @@ ln -s /Library/Frameworks/R.framework/Versions/3.6.x /Library/Frameworks/R.frame
 
 Then restart RStudio.
 
+## Issues with Fonts on MacOS
+
+On older versions of MacOS, it may happen that axis labels do not show up when calling `plot()` 
+(section "renaming factors" in "Starting with Data"). This issue might be due to the default font 
+Arial being deactivated, so that R cannot find it. To resolve this issue, go to Finder, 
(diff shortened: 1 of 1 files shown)


fmichonneau:
this is great, thank you!

fmichonneau:
just a note that the integration is currently broken because the R image on travis isn't as well curated as it used to be. So the changes I'm merging now will not be visible on the website until I get a chance to fix the issue. I opened https://github.com/datacarpentry/R-ecology-lesson/issues/652 to keep track of it.

PR #653, opened 2020-10-03
Use patchwork instead of gridExtra to arrange plots, improve example

Teebusch:
- patchwork has a much simpler syntax., is a tidyverse-adjacent package, and (arguably) the future of arranging ggplots
- added installation instructions for the package, fixed a code error  in the example (`labs()` was added twice to the 1st plot)  and reduced the complexity of the syntax to better highlight the important parts (shorter variable name, fewer linebreaks)
- gave "exporting plots" a separate headline, because it really is a separate section.
- the example now stacks plots vertically, which looks nicer (less crowded) when teaching using a low res screen / projector

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -587,36 +587,42 @@ ggplot(surveys_complete, aes(x = species_id, y = hindfoot_length)) +
 > * Try using a different color palette (see
 >   http://www.cookbook-r.com/Graphs/Colors_(ggplot2)/).
 
-## Arranging and exporting plots
+## Arranging plots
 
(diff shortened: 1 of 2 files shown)


mondorescue:
@Teebusch This is great!

@datacarpentry/ecology-curriculum-maintainers What do you think? Should we replace `gridExtra`? I think @Teebusch has a good point about `patchwork` having simpler syntax.

fmichonneau:
I thought we'd already done it here: https://github.com/datacarpentry/R-ecology-lesson/pull/596/files 

It must have been reverted inadvertently in a merge conflict resolution.

In any case, yes, I'm all in favor of this change. Thank you for doing this @Teebusch!

PR #654, opened 2020-10-03
closes #627, add missing font workaround for ggplot to instructor notes

### LINKED ISSUE #627: Resolution to issues of axes text not appearing when plotting with ggplot2
With teaching a recent workshop, we came across an issue where several individuals with old Macs could not get their axes text to show up when starting the ggplot2 lesson (no error was outputted, the plot axes just never showed up).  Our workshop coordinator mentioned that they had previously encountered this issue and believe that it is due to R not being able to find the default Arial font. 

To resolve this issue of no axes text, there are two approaches: (long way) Go to Finder Search for "Font Book" and open it. Look for the Arial font and, if it is greyed out, turn it on. This eliminates the warning and enables R to produce the appropriate labels.

(short way) Add `+ theme(text = element_text(family = "serif"))` to the ggplot2 code. 

Perhaps we could either add a note to the lesson or to the Instructor Notes?

Teebusch:
@rachellombardi reported a workaround for missing Arial font in ggplots on older Macs. I've added (slightly modified) instructions to the instructor notes.

### DIFF
--- a/instructor-notes.md
+++ b/instructor-notes.md
@@ -68,6 +68,11 @@ On older versions of MacOS, it may happen that axis labels do not show up when c
 Arial being deactivated, so that R cannot find it. To resolve this issue, go to Finder, 
 Search for Font Book and open it. Look for the Arial font and, if it is greyed out, turn it on. 
 
+If the problem occurs with `ggplot2` plots, an alternative workaround is to change the default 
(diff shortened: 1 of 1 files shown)


mondorescue:
Looks good to me. Thanks!

PR #655, opened 2020-10-03
acknowledge and explain warning in parsing dates when it occurs

Teebusch:
When parsing dates with lubridate, some fail, because two of the dates do not exist (2000-9-31 and 2000-4-31).
This isn't mentioned/explained in the lesson when the warning occurs, but only later. It may be good to acknowledge the warning right away, so no learners get confused about whether they did smething wrong.


fmichonneau:
thank you for clarifying this!

PR #656, opened 2020-10-03
add info on submitting issues to contributing.RMD

Teebusch:
- some contributors add "carpentries checkout contribution" or similar to their issue titles. This is unnecessary clutter in the issues list. I thought it might help to point out in the contribution guidelines that this is not needed.
- I also noticed that some issues would be very simple pull requests and would like to encourage contributors to commit their issues as pull requests. I find the Github Desktop app extremely helpful (and fun to use) and think it would be worth mentioning it.
- Some issues here are actually multipleissues in one. I suggest reminding contributors to open one issue per problem.

### DIFF
--- a/CONTRIBUTING.Rmd
+++ b/CONTRIBUTING.Rmd
@@ -23,7 +23,11 @@ our
 ### Submitting Issues on GitHub
 
 If you have an idea about how to improve the lesson, you can submit it as an 
-issue on GitHub. Submitting an issue can count as a contribution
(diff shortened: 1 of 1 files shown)


fmichonneau:
I'm OK with these changes. What do the other @datacarpentry/r-ecology-lesson-maintainers think?

mondorescue:
@fmichonneau I second!

fmichonneau:
Thank you for these changes @Teebusch

PR #664, opened 2020-10-20
Add info about R 4.0 to instructor notes and RSQLite package version to "before we start"

### LINKED ISSUE #640: Anything about versions needed?
What versions of packages does this (and other) lesson(s) work with?

Teebusch:
It's worthwhile to point out in the instructor notes that R <4.0.0 might cause problems.
I've also added the verison info of RSQlite to the `sessionInfo()` in 'before we start.
This probably closes #640


fmichonneau:
thank you for adding these clarifications.

Issue #612, opened 2020-05-26
Screen Reader Doesn't Read Learning Objectives  


klbarnes20:
Hi everyone, 

I am going through this lesson to learn about using R and I stumbled on a potential issue. I often use a screen reader when reading a lot online. I discovered that for this curriculum, my screen reader doesn't recognize the Learning Objectives in the gray box at the top of the lessons. I am using this chrome extension: https://chrome.google.com/webstore/detail/read-aloud-a-text-to-spee/hdhinadidafjejdhmfkjgnolgimiaplp?hl=en 

I'm not sure if this problem is unique to my extension or a broader problem, but I thought I would bring it up as a potential accessibility issue. 

Thanks!

ErinBecker:
Thank you for this issue @klbarnes20. @zkamvar and @fmichonneau are working on a new template prototype and including an accessibility review in the process. I'm pinging them here so they take this issue into consideration during the design phase.

fmichonneau:
@klbarnes20 that's interesting. It looks like this extension recognizes the learning objective in some episodes but not others. It'd be interesting to investigate a little further but I don't see anything specific in the source that would explain the difference.

Teebusch:
Using the extension, it seems that the objectives are only read for "Before we Start" and "Manipulating Data". The learning objects themselves are indeed always formatted alike (as far as I can tell). To me it seems like the issue may be with different levels of headings being used throughout the document, but I can't pinpoint it perfectly.

PR #671, opened 2020-11-05
Cleanup Rmds, edit navbar, and smaller fixes

### LINKED ISSUE #614: .org website layout issue.
Whats up I just wanted to let you know your datacarpentry.org header overlaps the body of your document. I'm on firefox 77.0.
dataCarp

Teebusch:
-  Use **custom block syntax (`:::`)** for objectives, challenges, and notes. 
    + This makes them easier to edit, because syntax highlighting and code execution in RStudio works normally (unlike when using blockquote syntax `>`)
    + It is semantically cleaner, because objectives are not quotes
    + It allows to style them separately
    + disadvantage: the blocks are now somewhat more difficult to spot in the code
- **CSS updates**
    + style challenges, objectives, notes and answers using the [carpentry lesson template colors](https://carpentries.github.io/lesson-example/04-formatting/index.html)
    + some other (minor) adjustments to the CSS
     + Note: The site loads CSS from different sources and I'm not a CSS wizard, so it wasn't always possible to do exactly what I wanted. I think it's an overall improvement, but the site's CSS could use some more work.
- **Revised headings** across all `.Rmd` files. There were some "illegal" 1st level headings, and some inconsistencies. I had hoped that this would fix the screen reader issue (#612), but unfortunately the issue persists.
- **Updated the navbar,** so that episodes are all in one dropdown menu now, and code of conduct, license, reference and instructor notes are linked. This confirms with the [lesson template](https://carpentries.github.io/lesson-example/index.html). It also minimizes use of space, so that it closes #614 
- **Updated the non-lesson `.Rmd` files** (license, instructor notes, Code of Conduct, Reference) so that they render more nicely. 
- **Updated chunk options** throughout the lesson:
    + Removed redundancies (mostly `purl=FALSE`), 
    + removed the default `results='hide'` and so that code chunk output now renders by default. I've added explicit `eval=FALSE` or `results='show'` where equired. Reason: not showing results is non-default behavior and likely confusing to contributors who ar not familiar with the repo (I know it was to me). In most cases where the output should not be shown, the computation is also not needed, so that it's arguably better to use `eval=FALSE`. If the computation is needed, but the output not, an explicit `results='hide'` is better.
    + chunk options are now uniform with no spaces around the `=`
- **Replaced Rarkdown figure syntax** `` with code chunks with `knitr::include_graphic()`. This is the preferred way to include graphics in RMarkdown (for reasons I can't remember). I may have missed some figures though.
- Some smaller fixes along the way.

Many of these changes, don't have a visible effect on the rendered lesson. However, I think it's important to keep the Rmd files clean and consistent to make it easier for people to contribute, and to (eventually)smoothen the transition to the official carpentries lesson template. 

I've tried my best to make sure everything works and looks as intended, but the PR should be reviewed thoroughly. Especially changing the chunk options is error prone. It should be checked whether all chunks render/don't render as desired.


ErinBecker:
Thank you for all of this work @Teebusch. I'm helping the current lesson Maintainers process outstanding pull requests on this repository, in preparation for [transition to the new lesson infrastructure](https://carpentries.github.io/workbench/faq.html#update-fork-from-styles). I've consulted with The Carpentries Curriculum Team and been advised that most or all of the changes introduced in this PR will be handled automatically by the [`sandpaper` package](https://github.com/carpentries/sandpaper) when this lesson is processed for transition to the workbench, which will occur next week. Because of that I am closing this PR without merging. Thank you again for your contribution.

PR #672, opened 2020-12-02
Do not refer to "select" for dplyr::filter

zkamvar:
The verb "Select" should not be used to refer to an operation made by `dplyr::filter()` because these two functions are often mixed up. I've proposed an alternative.


fmichonneau:
thank you, good catch.

PR #681, opened 2021-01-27
Missing values in data added to Starting with data lesson in R Ecology

sanjayfuloria:
I have added a section on missing values in data. A short write-up on missing values and how to handle those. In addition, I have shown how to perform list wise deletion using na.omit.

---

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -13,6 +13,7 @@ source("setup.R")
 > ### Learning Objectives
 >
 > * Load external data from a .csv file into a data frame.
+> * Handling missing values
 > * Describe what a data frame is.
 > * Summarize the contents of a data frame.
 > * Use indexing to subset specific portions of data frames. 
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
 view(surveys)
 ```
 
-> ### Note
->
+## Missing values in data
+Missing values are those not stored for a variable in the observation of interest. 
+This is a common problem across all domains of research. The reasons could be many. 
+Missing values may have a significant effect on the conclusions of the research. 
+When viewing above data, we observe a lot of cells showing ‘NA’ values. 
+Those are missing values. While care should be taken to collect data so that there 
+are no missing values, most of the times values go missing. 
+
+One of the ways to handle this is to omit those cases with the missing values. 
+Though, this would lead to a smaller sample size, it is unavoidable. 
+This approach is called complete case analysis or listwise deletion. 
+Let us do that using the following code.
+
+```{r, purl=FALSE}
+surveys <- na.omit(surveys)
+```
+## Checking the dataset and continuing with the analysis
+Now, when we check the dimensions of the dataset, we find 30,738 rows as against 34,786 rows earlier. 
+So, we have lost 4,048 rows. Let us continue with the analysis with this new dataset.
+
+>### Note
 > `read_csv()` assumes that fields are delineated by commas, however, in several
 > countries, the comma is used as a decimal separator and the semicolon (;) is
 > used as a field delineator. If you want to read in this type of files in R,


fmichonneau:
Thanks for your contribution @sanjayfuloria!

Overwriting the full `surveys` dataset by removing the missing data would have consequences for the rest of the lesson and such change should be evaluated carefully.

There is already a section on missing data on the previous chapter: https://datacarpentry.org/R-ecology-lesson/01-intro-to-r.html#missing-data May I suggest you propose a change to improve this existing section? 

What do the other @datacarpentry/r-ecology-lesson-maintainers think?

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
+Those are missing values. While care should be taken to collect data so that there 
+are no missing values, most of the times values go missing. 
+
+One of the ways to handle this is to omit those cases with the missing values. 
+Though, this would lead to a smaller sample size, it is unavoidable. 
+This approach is called complete case analysis or listwise deletion. 
+Let us do that using the following code.
+
+```{r, purl=FALSE}
+surveys <- na.omit(surveys)
```
Teebusch, line 154: I agree with @fmichonneau that we shouldn't overwrite the data frame here. It will affect later parts of the lesson. See for example `starting with data > renaming factors` where the missing are turned into "undetermined".

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
+```
+## Checking the dataset and continuing with the analysis
+Now, when we check the dimensions of the dataset, we find 30,738 rows as against 34,786 rows earlier. 
+So, we have lost 4,048 rows. Let us continue with the analysis with this new dataset.
```
Teebusch, line 158: Instead, one could say something like "If we do this, we would loose 4,048 rows. Often there are better ways of dealing with missing data than listwise deletion, so we will keep these missing values around for now."

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
+
+>### Note
```
Teebusch, line 160: There should be a space between `>#`

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
+## Missing values in data
+Missing values are those not stored for a variable in the observation of interest. 
+This is a common problem across all domains of research. The reasons could be many. 
+Missing values may have a significant effect on the conclusions of the research. 
+When viewing above data, we observe a lot of cells showing ‘NA’ values. 
+Those are missing values. While care should be taken to collect data so that there 
+are no missing values, most of the times values go missing. 
+
+One of the ways to handle this is to omit those cases with the missing values. 
+Though, this would lead to a smaller sample size, it is unavoidable.
```
Teebusch, line 149: I'm not sure what you mean with "It is unavoidable". There are imputation techniques and modelling approaches (e.g. FIML) that can deal with missing data without reducing the sample size.

02-starting-with-data.Rmd
```diff
@@ -13,6 +13,7 @@ source("setup.R")
 > ### Learning Objectives
 >
 > * Load external data from a .csv file into a data frame.
+> * Handling missing values
```
Teebusch, line 16: I'm new to this, but I believe changes to lesson objectives should be approved by CAC (Curriculum Advisory Committee)?

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
 
-> ### Note
->
+## Missing values in data
+Missing values are those not stored for a variable in the observation of interest. 
+This is a common problem across all domains of research. The reasons could be many. 
+Missing values may have a significant effect on the conclusions of the research. 
+When viewing above data, we observe a lot of cells showing ‘NA’ values. 
+Those are missing values. While care should be taken to collect data so that there 
+are no missing values, most of the times values go missing.
```
Teebusch, line 146: I'd argue that values rarely "go missing". Usually the data can't be collected in the first place.

02-starting-with-data.Rmd
```diff
@@ -136,8 +137,27 @@ To open the dataset in RStudio's Data Viewer, use the `view()` function:
...
->
+## Missing values in data
+Missing values are those not stored for a variable in the observation of interest. 
+This is a common problem across all domains of research. The reasons could be many. 
+Missing values may have a significant effect on the conclusions of the research. 
+When viewing above data, we observe a lot of cells showing ‘NA’ values. 
+Those are missing values. While care should be taken to collect data so that there 
+are no missing values, most of the times values go missing. 
+
+One of the ways to handle this is to omit those cases with the missing values.
```
Teebusch, line 148: It think it's a bit unusual to do listwise deletion at this stage of the data analysis process and for *all* variables. Normally this would be done once we know which variables are used in the analysis and only if the missingness is a problem.

Teebusch (changes requested):
Thank you for your contribution. I would suggest some changes, Please see my review.

Teebusch:
> There is already a section on missing data on the previous chapter: https://datacarpentry.org/R-ecology-lesson/01-intro-to-r.html#missing-data May I suggest you propose a change to improve this existing section?
 

I agree

sanjayfuloria:
Hi! Francois,

Thanks for your reply. I would suggest adding my inputs as an optional
exercise in this lesson. Dealing with missing data is an important aspect
of any analysis. One way to deal with this is to use case wise deletion.
After this exercise, participants could continue with the original (with
missing values) dataset. This will resolve the problem of modifying other
parts of the lesson. I would be happy to hear about your response to this.

Thanks and regards,
Sanjay

mondorescue:
> > There is already a section on missing data on the previous chapter: https://datacarpentry.org/R-ecology-lesson/01-intro-to-r.html#missing-data May I suggest you propose a change to improve this existing section?
> 
> I agree

@fmichonneau I support this, too.

ErinBecker:
Thank you @sanjayfuloria for your contribution. I'm helping the current lesson Maintainers process outstanding pull requests on this repository, in preparation for [transition to the new lesson infrastructure](https://carpentries.github.io/workbench/faq.html#update-fork-from-styles). Based on the comments above from @fmichonneau, @Teebusch, and @mondorescue, I am closing this PR without merging.

Issue #680, opened 2021-01-26
Subsetting tibbles with `df[, col]` returns a data frame, not a vector


mikemahoney218:
There appears to be an error in the [indexing and subsetting data frames section of Starting with Data](https://datacarpentry.org/R-ecology-lesson/02-starting-with-data.html#Indexing_and_subsetting_data_frames). While subsetting base data frames with `df[, col]` returns a vector, subsetting tibbles with this format return a dataframe identically to subsetting via `df[col]`:

``` r
class(iris[, "Species"])
#> [1] "factor"
class(tibble::as_tibble(iris)[, "Species"])
#> [1] "tbl_df"     "tbl"        "data.frame"
```

<sup>Created on 2021-01-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I imagine this has changed since the lesson was first written as tibble strives to improve consistency. As we read in the `surveys` data set with `readr::read_csv`, the table is read in as a tibble, and the comment in the text (that subsetting with the comma returns a vector) is incorrect. 

With that said, I don't know if it makes sense to specifically highlight that this is one of the differences with tibbles that the lesson earlier says isn't worth getting into, or to not mention that base data frames behave differently. It might make sense to mention when introducing tibbles that the goal of tibbles are to provide more consistent behavior than base data frames, which would provide a basis for flagging those differences as they come up?

fmichonneau:
Good catch! With the transition to using tibbles, I think this section could be greatly simplified. I don't think it's necessary to get into the details of the differences between tibbles and data.frames given the intended audience for these lessons.

Would you be willing to start a pull request that addresses this issue?

Thank you!

mikemahoney218:
Of course! Opened #683 .

Teebusch:
closed by #683

PR #683, opened 2021-02-05
Update subsetting for tibbles (#680)

mikemahoney218:
This is my attempt at rewriting the section around subsetting to consider the more consistent behavior of tibbles, which is what our novices will be principally using. 

There is a call-out that this behavior should only be expected when working with tibbles on line 289. I agree that it's not worth diving into the differences between base data frames and tibbles at this level, but I do want to at least flag "this is not universally true." If this feels likely to confuse, though, happy to remove it.

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -254,22 +254,47 @@ want from it. Row numbers come first, followed by column numbers. However, note
 that different ways of specifying these coordinates lead to results with
 different classes.
 
-
 ```{r, purl=FALSE}
-# first element in the first column of the data frame (as a vector)
-surveys[1, 1]   
-# first element in the 6th column (as a vector)
+# We can extract specific values by specifying row and column indices
+# in the format: 
+# data_frame[row_index, column_index]
+# For instance, to extract the first row and column from surveys:
+surveys[1, 1]
+
+# First row, sixth column:
 surveys[1, 6]   
-# first column of the data frame (as a vector)
-surveys[, 1]    
-# first column of the data frame (as a data.frame)
-surveys[1]      
-# first three rows of the 6th column (as a vector)
-surveys[1:3, 6] 
-# the 3rd row of the data frame (as a data.frame)
-surveys[3, ]    
-# equivalent to head_surveys <- head(surveys)
-head_surveys <- surveys[1:6, ] 
+
+# We can also use shortcuts to select a number of rows or columns at once
+# To select all columns, leave the column index blank
+# For instance, to select the all columns for the first row:
+surveys[1, ]
+
+# The same shortcut works for rows --
+# To select the first column across all rows:
+surveys[, 1]
+
+# An even shorter way to select first column across all rows:
+surveys[1] # No comma! 
+
+# To select multiple rows or columns, use vectors!
+# To select the first three rows of the 5th and 6th column
+surveys[c(1, 2, 3), c(5, 6)] 
+
+# We can use the : operator to create those vectors for us:
+surveys[1:3, 5:6] 
+
+# This is equivalent to head_surveys <- head(surveys)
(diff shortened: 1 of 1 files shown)


02-starting-with-data.Rmd
```diff
@@ -254,22 +254,47 @@ want from it. Row numbers come first, followed by column numbers. However, note
...
-# first three rows of the 6th column (as a vector)
-surveys[1:3, 6] 
-# the 3rd row of the data frame (as a data.frame)
-surveys[3, ]    
-# equivalent to head_surveys <- head(surveys)
-head_surveys <- surveys[1:6, ] 
+
+# We can also use shortcuts to select a number of rows or columns at once
+# To select all columns, leave the column index blank
+# For instance, to select the all columns for the first row:
```
Teebusch, line 269: Very nice explanation!

02-starting-with-data.Rmd
```diff
@@ -278,17 +303,23 @@ or decreasing order, test `1:10` and `10:1` for instance.
 You can also exclude certain indices of a data frame using the "`-`" sign:
 
 ```{r, purl=FALSE}
-surveys[, -1]          # The whole data frame, except the first column
+surveys[, -1]         # The whole data frame, except the first column
 surveys[-(7:34786), ] # Equivalent to head(surveys)
```
Teebusch, line 307: ...or `tail(34779)`, I suppose  :)

Teebusch (approved):
Looks excellent!

(I think it's ok to talk about data frames rather than tibbles. We mention the difference briefly somewhere in the lesson nd say that it's not important.)

02-starting-with-data.Rmd
mikemahoney218: I tried my best to keep to a minimal rewrite :joy: 

If anything, it almost makes sense to move this one a little later in the lesson so that we can subset using `surveys[-7:nrow(surveys), ]` rather than hard-code the number of rows in the table, which feels a _little_ bit like teaching people to use "magic numbers" instead of semantic functions or variables. But this feels like a less clear case than the rest of the PR, so I just left it as it was :joy:

Teebusch: Very good point about not using "magic numbers". I believe `nrow()` is mentioned earlier in this lesson (`inspecting data frames`), so it would be ok to use it here, wouldn't it?

I know you want to keep the rewrite minimal, but from my perspective using `nrow()` instead of `34786` is much better. And since you are already improving this section, you could just fix this along the way. 
 @fmichonneau , what do you think?

mikemahoney218:
@Teebusch @fmichonneau I just pushed a commit with `nrow` replacing the "magic number" in the subsetting example, as we discussed last month -- not sure if it needs another review or not!

mikemahoney218:
CI failed with `unable to load shared object '/home/runner/work/_temp/Library/stringi/libs/stringi.so'` which seems to be unrelated to these changes (all commits have failed in past few days).

Teebusch (approved):

02-starting-with-data.Rmd
Teebusch: I just noticed a typo: "to select ~the~ all columns for the first row:"

Teebusch:
fixed a typo, minor changes to comments

Teebusch (approved):

Teebusch:
I believe this PR can be merged. I'll leave it to one of the other maintainers to have a final look and push the button. :)

fmichonneau:
This looks good! Thank you both for your work on this. I think introducing the double bracket might be a little beyond the scope of this lesson for the intended audience, but it could be useful in some contexts. Ideally, we'd want to add a mention of this in the instructor notes. Happy to merge as is though.

Issue #685, opened 2021-03-02
Minor suggestion to mention View() when presenting view()


paulapappalardo:
Hi. I wonder if in the section "Starting with data" it could be useful to mention that we are using *view* (lowercase v)  from the package tibble (under the umbrella of tidyverse), but that *View* (uppercase v) would also work to view data frames because is part of base R?. We talked earlier in the episode about the functions from specific packages versus the functions that come built into R, so it shouldn't be a new concept. I have always used *View* and just discovered *view* myself when looking through the  R-ecology-lesson :)

mondorescue:
I think you have a good point, especially since `view()` in the **tibble** package is considered *experimental* by Müller, et al. Would you be willing to provide a suggested edit (i.e. Pull Request) to the text?

paulapappalardo:
Sounds good, I will do a Pull Request (my first PR, so reading a bit to be sure I do not do something wrong). It seems I need to select or __base: main__ or __compare: main__, which one is the appropriate choice?

mondorescue:
@paulapappalardo thanks for making the edits. Reviewing the changes, it seems that the note about `view()` vs `View()` may be better placed in the [Instructor Notes](https://github.com/datacarpentry/R-ecology-lesson/blob/main/instructor-notes.md) because the call out box is about `read_csv()`. Also (now thinking about it), `view()` vs `View()` isn't about importing data in R but rather about a "feature" of RStudio. What do you think?

paulapappalardo:
That sounds good! And the instructor can decide if to mention it or at least have it in mind in case anybody asks.

mondorescue:
@paulapappalardo Thanks for the contribution!

PR #696, opened 2021-03-24
Minor change to address issue #685

paulapappalardo:
I added a few lines to the 02-starting-with-data lesson to clarify that the base function `View()` can also be used to view a data frame.


mondorescue:
@paulapappalardo if you don't mind resubmitting a PR with edits made to the instructors note instead (as discussed in #685), that will be great. I'll close this PR.

PR #705, opened 2021-04-13
Add note about ggplot2 vs. ggplot

katbeescience:


fmichonneau:
thanks for this addition!

PR #709, opened 2021-04-16
fix-genus-species-confusion

### LINKED ISSUE #706: Should use genus rather than species in description in the "Visualizing data" section
I found a few minor discrepancies in the use of species vs genus in the "Visualizing data" section. Specifically, there are two instances where the genus is being plotted, but the description states that the learner is distinguishing species. 

- The first instance is under "Plotting time series data" where the description states "We will be able to distinguish species in the plot if we add colors" but the code uses "color = genus". This means that the description should be changed to "We will be able to distinguish genera in the plot if we add colors" [line 352 in 04-visualization-ggplot2.Rmd]. 
- The second instance is under "Faceting" where the description states "We will use it to make a time series plot for each species" but again the code uses "facet_wrap(facets = vars(genus))". So the description should be "We will use it to make a time series plot for each genus" [line 390-391 in 04-visualization-ggplot2.Rmd]

---

Teebusch:
closes #706

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -349,7 +349,7 @@ ggplot(data = yearly_counts, aes(x = year, y = n, group = genus)) +
     geom_line()
 ```
 
-We will be able to distinguish species in the plot if we add colors (using 
+We will be able to distinguish genera in the plot if we add colors (using 
 `color` also automatically groups the data):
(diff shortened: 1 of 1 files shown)


fmichonneau:
thank you for fixing this @Teebusch and thank you @rmgiglio for pointing this inconsistency.

PR #711, opened 2021-04-21
Fixes #518, using suggestion 2 from  @kimh11

### LINKED ISSUE #518: Suggestion: change column "genus" to "genera" either in dataset or with rename() function
I suggest either adding a short code piece to use the rename() function to change the column "genus" to "genera" (thus alerting the learners to their relationship here, while adding a new function) or changing the column name in the original dataset.  Otherwise, I've found that using the correct plural for genus confuses learners who are not biologists.  Although it's the R ecology lesson and one might assume biology knowledge, at my institution we use it for learners from multiple disciplines who are not familiar with the plural genera.

Teebusch:
This closes #518 by adding clarification that genera is the plural of genus. Personally, I don't think this change is necessary, as this is clear from the current version of the challenge text ("how many different genera are in the genus column?"). I would also be fine with rejecting this PR and closing #518 without any changes.


Teebusch:
Oh sorry, I messed up this PR. I will make a new one.

PR #712, opened 2021-04-21
add clarification that genera is the plural of genus, closes #518

### LINKED ISSUE #518: Suggestion: change column "genus" to "genera" either in dataset or with rename() function

Teebusch:
This closes #518 using suggestion number 2, adding clarification to the challenge text.

Personally, I think it's clear enough from the current challenge text ("How many different genera are in the `genus` column?") so I would be ok with rejecting this PR and closing #518.

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -461,7 +461,7 @@ this information built in. It is particularly helpful when there are many levels
 > 2. Using the functions you learned before, can you find out...
 >
 >      * How many rabbits were observed?
->      * How many different genera are in the `genus` column?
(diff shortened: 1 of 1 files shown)


Teebusch:
I will reject and clode this, because #518 is closed and I think the current text is clear enough.

Issue #718, opened 2021-05-09
Missing answers to challenges


DrMaggie:
Hi, 

My apologies in advance if creating a new issue "on top of" an existing but now closed one is against github "netiquette", but I'm quite new to this platform! 

Very similar to the now closed issue #717 (https://github.com/datacarpentry/R-ecology-lesson/issues/717), I am experiencing pretty much the same problem: when visiting the part of the web page https://datacarpentry.org/R-ecology-lesson/04-visualization-ggplot2.html that describes the ggplot challenge to customize  violin & box plots, there is no  clickable "button" to expand & show the solution code. I attach a screen shot of the relevant part of the page as displayed in my browser:
![screenshot](https://user-images.githubusercontent.com/8366663/117573962-292be100-b0db-11eb-8d3d-44ad6ceccb89.jpg)

Maybe the fix that was applied before the 717 issue was closed somehow doesn't work for all browsers? (I use the latest version of Chrome for Windows 10 (Version 90.0.4430.93, official 64-bit build)...)

Best regards,
Maggie

Teebusch:
Hi @DrMaggie , thank you for bringing this up. To me it looks like for this particular challenge there is no answer provided in the code, so the problem is not that the expand button is broken (#717), but that there is nothing to expand (new issue).

To solve this we would have to add an answer block -- somthing like this:

```r
> ```{r boxplot-challenge, answer=TRUE, purl=FALSE}
> # violinplot
> ggplot(data = surveys_complete, mapping = aes(x = species_id, y = weight)) +
>     geom_violin() +
>    geom_jitter(alpha = 0.3, color = "tomato") +
>    scale_y_log10()
> 
> # hindfoot length boxplot
> ggplot(mapping = aes(x = species_id, y = hindfoot_length)) +
>    geom_boxplot(alpha = 0) +
>   geom_jitter(mapping = aes(color = as_factor(plot_id)))
>```
```

I don't have time to test this and make a PR right now. In particular, I'm not sure the `as_factor()` is even necassary, since we moved to tidyverse's `read_csv()`. If it isn't, then that part should be removed from the challenge.

DrMaggie:
Dear @Teebusch ,

thank you so much for addressing my question, and for providing the solution code - it is very much appreciated, and really helped me to better understand the possibilities of ggplot!

[...]

PR #724, opened 2021-05-31
add challenge about boxplot outlier plotting

haasek:
I added a question to the challenges after the boxplot section that asks learners to reflect on the alpha parameter and its effects. It doesn't get clarified elsewhere that setting alpha to zero for the boxplot layer leads to the removal of the outlier points.
When we taught the material, we noticed that the parameter often got removed during the exercises, which leads to double-plotting of the outliers, once in the boxplot and once in the jitter layer.
I thought it could be helpful to make it part of the challenge to understand this behaviour and find a way of combining a jitter and a boxplot layer without duplicated outliers and without changing the transparency.

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -294,6 +294,10 @@ hidden?
 > - Create boxplot for `hindfoot_length`.  Overlay the boxplot layer on a jitter
 >   layer to show actual measurements.
 >
+> - What effect does the `alpha` parameter in the boxplot layer have on the outlier
+>   points? Can you achieve the same visualisation without changing the transparency?
+>   Hint: Check the `geom_boxplot` parameters starting with `outlier...`.
+>
 > - Add color to the data points on your boxplot according to the plot from which
 >   the sample was taken (`plot_id`).
 


Teebusch:
Thank you for the PR! I agree with your point -- I think when I was teaching this recently, I also explained the problem with duplicating outliers and how to remove the outliers in the box+jitter plot. This should definitely be adressed in the lesson.

(Another thing, although maybe not necessary to fix in this PR: the challenge does not have an answer and is complicated enough that it would benefit form having one, I think. This is related to #718)

haasek:
Yes, definitely. I was unsure if I should just add a solution to the outlier issue or the whole block of challenges. If you want, I can give it a go and add answers for the whole block.

ErinBecker:
I'm helping the current lesson Maintainers process outstanding pull requests on this repository, in preparation for [transition to the new lesson infrastructure](https://carpentries.github.io/workbench/faq.html#update-fork-from-styles). 

Thank you for this contribution @haasek. With https://github.com/datacarpentry/R-ecology-lesson/pull/817, `outlier.shape = NA` has replaced `alpha = 0` in the lesson text. There seems to be general agreement that this update makes the lesson clearer for learners! Thank you for contributing to this discussion.

PR #725, opened 2021-06-11
Match challenge answer to prompt & style fix

jessicaguo:
The challenge starting in line 374 requests the number of rodents observed in each plot, in each year. However, the dplyr syntax answer only groups by taxa and year. I have added 'plot_id' to line 398, so that the output matches the prompt. 

In addition, the code chunks starting in line 396 and 449 use 'tally', which I replaced with 'tally()' to match the tidyverse style guide.


fmichonneau (approved):

fmichonneau:
Thanks for this!

PR #727, opened 2021-06-15
Added in current units

### LINKED ISSUE #726: dplyr - mutate challenge: reminder of current units
Referring to the dplyr – mutate section, the mutate challenge (currently line 264) requests a new column for hindfoot_length_cm. The original data in surveys comes from the prior section (starting with data) which included the definitions of each column, where hindfoot_length is listed in mm. As this information was some time ago, it might be a useful reminder of the current units of hindfoot_length in this challenge. 

“Create a new data frame from the `surveys` data that meets the following criteria: contains only the `species_id` column and a new column called `hindfoot_cm` containing the `hindfoot_length` values (currently in mm) converted to centimeters. In this `hindfoot_cm` column, there are no `NA`s and all values are less than 3.”

sstevens2:
We noticed this in the workshop I taught yesterday.  @sawildman filed an issue.  Think this PR should implement the change he suggested.

based on issue #726 from @sawildman

Closes #726

### DIFF
--- a/03-dplyr.Rmd
+++ b/03-dplyr.Rmd
@@ -265,7 +265,8 @@ symbol negates the result, so we're asking for every row where weight *is not* a
 >
 >  Create a new data frame from the `surveys` data that meets the following
 >  criteria: contains only the `species_id` column and a new column called
->  `hindfoot_cm` containing the `hindfoot_length` values converted to centimeters.
+>  `hindfoot_cm` containing the `hindfoot_length` values (currently in mm)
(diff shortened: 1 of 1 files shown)


sstevens2:
To put some motivation here.  I couldn't remember the units and someone asked. Think it would be good to add a reminder in the exercise itself.

Teebusch (approved):
Thank you! This looks good to me and is a useful addition.

fmichonneau (approved):

PR #737, opened 2021-07-05
guidelines for managing package library, closes #240

### LINKED ISSUE #240: Provide guidelines for better managing package library
A common reason people don't want to upgrade their R version, is that they need to reinstall all their packages afterwards.

There are several workarounds this issue, but they vary across OSes and can be a little technical, but it seems that having a guide somewhere that explains strategies to keep your packages across upgrades would be useful to our learners and the general public.

This is not urgent though, and should be addressed after the April lesson release.

(tangentially related to #234)

Teebusch:
This would close #240

- I wonder if the Preparations section is the  right place for this information or if this would fit better in "Before we start", maybe before "organizing your working directory"
- Is it too detailed? 
- is anything missing?

### DIFF
--- a/index.Rmd
+++ b/index.Rmd
@@ -110,7 +110,7 @@ your operating system below. You have to install R before you install RStudio.
 
 ### Update R and RStudio
 
-If you already have R and RStudio installed, check if your R and RStudio are 
+If you already have R and RStudio installed, first check if your R version is
 up to date:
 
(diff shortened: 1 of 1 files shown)


fmichonneau (approved):

fmichonneau:
This looks good to me! Thanks.

Issue #606, opened 2020-04-13
Add challenge to 'Formatting Dates' section of 'Starting with Data'


alexis-catherine:
The only section of the Starting with Data section is the Formatting Dates. One thought is a challenge could be to re-write the ymd() code with one of ymd() sister functions. It would help demonstrate how flexible the different functions are and how if your data is in a slightly different format it isn't an insurmountable problem.

Challenge: 
1. Different countries and organizations have different ways they represent dates month-day-year, day-month-year, and year-day-month are all common variants. ymd() is part of a family of functions that can parse dates in different formats. Use the help function or the help operator to read through the documentation, pick one of the sister functions, and rewrite the code to parse the dates using that function instead.

GitHubDoug:
A common issue is plotting data versus elapsed, vs. absolute, time or date.
A further challenge suggestion:
2.  Generate a new column, days_elapsed, by subtracting the lowest value of surveys$date from all other values of surveys$date.
Hint:  Base R function min(c(1, 2, 0, NA) , na.rm) gives the minimum value of the vector c(1, 2, 0, NA)

Answer:
surveys$days_elapsed <- surveys$date - min(surveys$date, na.rm = TRUE)
str(surveys$days_elapsed)

A common error would be assuming that surveys$date[1] is the minimum value, or the units of the output, which in this case are days.

tobyhodges:
Thanks both for contributing to this discussion. The lesson underwent a major update and reorganisation when https://github.com/datacarpentry/R-ecology-lesson/pull/887 was merged. As this issue relates to content in a version of the lesson before that update took place, I will close it. Please open a new issue if you believe that some or all of the changes being discussed here are remain relevant to the redesigned lesson, linking to this thread where appropriate.

PR #740, opened 2021-07-09
modification to working with dates section

### LINKED ISSUE #678: formatting dates section
In the formatting dates section, it will be useful to mention the other functions that be used to have a different date format like  `dmy()`, `ydm()` etc. 
As a lot of people have difficulty with dates and most dates are not stored in separate column could be worth showing an example with `as.date()` function.

### LINKED ISSUE #730: Suggest csv editor in section on dates?
Because of the date conversions excel makes, would it be useful to suggest an alternative csv editor in the section "Formatting dates" in Episode 2: Starting with data?  This idea came up in instructor training and I agree it would be helpful to suggest a few alternatives.  I personally am looking for a good windows one, so I would have really appreciated the suggestion myself.

Teebusch:
Minor changes to the "Working with dates section". 

closes #730
closes #678

**Details:**

- Remove "As recommended earlier in this lesson, dates should be stored with one component per column". This wasn't  recommended earlier in the lesson and I don't think this way to store dates should be recommended. I believe this closes #730. The other issue brought up in 730 (recommend CSV editor) has been settled by discussion.
- mention other lubridate functions . I think this closes #678. I kept it more general than suggested by @ramay If we want to introduce another specific lubridate function, we could do so as part of a new challenge (see #606)

### DIFF
--- a/02-starting-with-data.Rmd
+++ b/02-starting-with-data.Rmd
@@ -682,11 +682,16 @@ Start by loading the required package:
 library(lubridate)
 ```
 
-`ymd()` takes a vector representing year, month, and day, and converts it to a
-`Date` vector. `Date` is a class of data recognized by R as being a date and can
+The **`lubridate`** package has many useful functions for working with dates. 
+These can help you extract dates from different string representations, 
+convert between timezones, calculate time differences and more. You can find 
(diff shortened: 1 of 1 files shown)


fmichonneau:
perfect, thanks!

The "as recommended earlier..." parts comes from the spreadsheet lesson that starts the workshop but I don't think your fix addresses this well.

PR #741, opened 2021-07-09
Fix Exporting plots code

### LINKED ISSUE #738: replace `grid.arrange` with `patchwork` in exporting plots section
This comment from @DrMaggie was buried in another issue, but this issue is actually a leftover from  a previous change where we replaced `{gridExtra}` with `{patchwork}`. The section should be adapted so that patchwork is used instead of gridExtra.

> Reading through all the comments above was very interesting and enlightening, not the least for a novice instructor as myself! I note that in addition to streamlining the narrative (and the reasoning behind what type of plot is useful for different purposes), it is also important to have clear examples that do not simultaneously introduce new concepts _and_ add a level of complexity. 

> One example of the latter is the final example, under ["Exporting plots"](https://datacarpentry.org/R-ecology-lesson/04-visualization-ggplot2.html#Exporting_plots) - here, the code example suddenly introduces the grid.arrange() function without mentioning that this requires a further package to be installed, i.e. gridExtra and/or arrangeGrob. 

```r
## This also works for grid.arrange() plots
combo_plot <- grid.arrange(spp_weight_boxplot, spp_count_plot, ncol = 2, 
                           widths = c(4, 6))
ggsave("combo_plot_abun_weight.png", combo_plot, width = 10, dpi = 300)
```
> (At least on my system, with only tidyverse loaded, the above piece of code didn't execute, but after googling and installing gridExtra, I got some reasonable output ;-)

_Originally posted by @DrMaggie in https://github.com/datacarpentry/R-ecology-lesson/issues/93#issuecomment-838573018_

Teebusch:
Closes #738

There was leftover code from previous version, using  `grid.arrange()` to combine plots. This has now been replaced with patchwork syntax. 
I've also modified the exported plot and renamed variables in the section on patchwork, so that the two sections are more consistent.

### DIFF
--- a/04-visualization-ggplot2.Rmd
+++ b/04-visualization-ggplot2.Rmd
@@ -610,16 +610,16 @@ determine how much space each plot uses:
 ```{r patchwork-example, message = FALSE, purl = FALSE, fig.width = 10}
 library(patchwork)
 
-plot1 <- ggplot(data = surveys_complete, aes(x = species_id, y = weight)) +
+plot_weight <- ggplot(data = surveys_complete, aes(x = species_id, y = weight)) +
   geom_boxplot() +
   labs(x = "Species", y = expression(log[10](Weight))) +
   scale_y_log10()
 
-plot2 <- ggplot(data = yearly_counts, aes(x = year, y = n, color = genus)) +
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks @Teebusch!

PR #771, opened 2021-09-27
remove inaccurate description of expected output of read_csv().

### LINKED ISSUE #761: Output of read_csv()
In the [Starting with data](https://datacarpentry.org/R-ecology-lesson/02-starting-with-data.html) episode, the lesson tells us that after calling `surveys <- read_csv("data_raw/portal_data_joined.csv")` we will get a message that says `Parsed with column specification` followed by the column name and data type. However in the example (and in my own work) the output header is just `Column specification`.  This should be clarified so learners don't think there is something wrong.

Teebusch:
fixes #761


fmichonneau:
thank you for taking the time to put this PR in!

PR #777, opened 2021-12-08
fixed broken link

kurtshowmaker:
I fixed the broken link for the Rstudio cheatsheet to the published link that works.

This new location is the location referred to as published in the original  repos readme.


fmichonneau:
Thank you for fixing this!

PR #781, opened 2021-12-10
Added accessibility options to resolve #743

### LINKED ISSUE #743: Adding more info about RStudio options in "Knowing your way around RStudio"
These materials are really excellent, and I have a small suggestion that you should feel free to take or leave as you see fit. In the section "Knowing your way around RStudio" it might be beneficial to highlight the different options that users can change about the RStudio appearance in Tools > Global Options.

Some of these options available in "Global Options" are really helpful and if users know more about how to fiddle with these options then they might feel more comfortable in RStudio in the long-term. For instance, soft-wrapping lines of code can be really helpful, as can changing background/font colors and sizes (particularly for colorblind people or those with vision difficulties). One option I've found especially helpful when I've taught workshops is to have people change the pane layout so that the Source is on the upper left, the Console is on the upper right, and the Files/Plots/etc panel is on the bottom right. If the bottom panels are minimized, then there's more space allocated to the inputs (Source) and outputs (Console) that users can see on the screen.

bmle:
Per issue #743:
1. Listed out the panes to make it easier to read
2. Added a sentence about adjusting colors and size

### DIFF
--- a/00-before-we-start.Rmd
+++ b/00-before-we-start.Rmd
@@ -125,12 +125,19 @@ workshop.
 Environment/History, Files/Plots/Packages/Help/Viewer,
 Console.](img/rstudio-screenshot.png)
 
-RStudio is divided into 4 "Panes": the **Source** for your scripts and documents
-(top-left, in the default layout), your **Environment/History** (top-right)
-which shows all the objects in your working space (Environment) and your command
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thank you for this!

PR #793, opened 2022-05-11

ajank:
Make it explicit that we want the last row from `surveys`, not `surveys_200`.

(Taught this lesson last month, and some of the learners tried to work on `surveys_200` here; they noted that this object has only one row already.)


fmichonneau (approved):

fmichonneau:
Thank you @ajank

Issue #788, opened 2022-02-17
Clarifying `view()` and `View()` functions in 02-starting-with-data.Rmd


renikaul:
Starting at line 135 the lesson uses `view()` to open the data in RStudio's Data Viewer. 
I suggest adding to the note on line 141 that `tibble::view()` and `utils::View()` have similar functionality but are part of different packages which is why the function does not seem to be case sensitive.

PR #805, opened 2022-06-20
Added note about view() vs View()

susanwashko:
Added a note clarifying the use of view() vs View(). This fixes issue #788.

Also moved read_csv note closer to where the read_csv function was used in the text.


fmichonneau (approved):

fmichonneau:
Thank you for this change!

PR #846, opened 2023-05-12
Relocated an instructor note

doujouDC:
We moved the following section:

In lesson 2 starting with data, one might not have the appropriate folder "data\_raw" in their working directory causing an error. This is a good time to go over reading an error, and a brief introduction of how to identify your working directory `getwd()` as well as setting your working directory `setwd("/somedirectory")` and if needed creating a directory within your script `dir.create("/some_new_directory")`, or simply creating it within a file explorer works if short on time.

to lesson 2

### DIFF
--- a/episodes/02-starting-with-data.Rmd
+++ b/episodes/02-starting-with-data.Rmd
@@ -62,6 +62,15 @@ We created the folder that will store the downloaded data (`data_raw`) in the ch
 If you skipped that part, it may be a good idea to have a look now, to make
 sure your working directory is set up properly.
 
+::: instructor
+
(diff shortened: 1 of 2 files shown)


jamesmbaazam:
Hi Doug, thank you for the change. I am happy to merge this PR but would like some explanation about why this was necessary.

doujouDC:
Hi James,

Sorry that I did not document my change. To be honest I cannot remember why I asked for that section to me moved. I think I felt at the time that it should have occurred earlier in the lesson. However, if no one else has noticed it is probably not worth changing.

Thanks so much for helping to maintain these lessons.

Doug Joubert

jamesmbaazam:
It is a pleasure, Doug. I will close this PR now but in case you remember why this change was necessary, we can reopen it and merge it in.

PR #863, opened 2023-06-23
move setup materials to learners/setup.md

### LINKED ISSUE #844: Move setup instructions to learners/setup.md
The setup instructions currently are in index.md, but they should be moved to learners/setup.md

zkamvar:
This moves the setup materials to the correct place in the lesson.

Background: This lesson was built in a different way than any other lessons and
the original authors had placed the setup inside of the index file. During the
transition to The Workbench, we included a template setup document that would
normally be overwritten with the lesson's setup document. In this case, it did
not exist, so it was not overwritten and thus contained links that went nowhere.

I was contacted by private email that someone was having trouble finding the
links for the data set and code handouts and this pull request will fix that.

This will fix #844


index.Rmd
```diff
@@ -8,7 +8,7 @@ site: sandpaper::sandpaper_site
 
 <div style="text-align: center; margin-top: 30px; margin-bottom: 30px;">
 
-![](./fig/DC-logo-vision.png)
+![](episodes/fig/DC-logo-vision.png){alt=""}
```
zkamvar, line 11: Note: I also updated the link to point to the correct place and added `alt=""` to indicate that this is a decorative image.

jamesmbaazam:
Thanks, Zhian. I will rebase and merge this as I would like to believe all is in order.

PR #869, opened 2023-08-01
Restore code-handout.R to workbench lesson

wood-chris:
Step 5 in the `00-before-we-start` lesson says:

> Download the [code handout](https://datacarpentry.org/R-ecology-lesson/code-handout.R), place it in your working directory and rename it (e.g., data-carpentry-script.R).

The "code handout" link is to https://datacarpentry.org/R-ecology-lesson/code-handout.R, which is in the legacy/gh-pages branch but not in the current `main` branch. I have copied this file to `main`.

However, I do note that `main` _does_ contain a file called `make_code_handout.R` - it's not clear to me if this is the file that's meant to be downloaded instead (and if so, please reject my PR and the link can simply be fixed instead :) )

### DIFF
--- a/code-handout.R
+++ b/code-handout.R
@@ -1,0 +1,645 @@
+
+### Creating objects in R
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+### Challenge
+##
+## What are the values after each statement in the following?
+##
+## mass <- 47.5            # mass?
+## age  <- 122             # age?
+## mass <- mass * 2.0      # mass?
+## age  <- age - 20        # age?
+## mass_index <- mass/age  # mass_index?
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+### Vectors and data types
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+## ## We’ve seen that atomic vectors can be of type character, numeric, integer, and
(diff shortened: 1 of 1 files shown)


zkamvar:
Hi @wood-chris,

I apologise for the confusion. The solution is to fix the markdown text, not to add the code. 
The file is now at <https://datacarpentry.org/R-ecology-lesson/files/code-handout.R>

The Workbench generates the code handout automatically from the code that exists and so adding it in the main branch is not necessary. The `make_code_handout.R` file was written in a time when the tools to build the lesson lived inside this repository and should be removed. That was an oversight on my part during the lesson transition back in May. 

cc: @datacarpentry/r-ecology-lesson-maintainers and @datacarpentry/r-ecology-lesson-maintainers-workbench

wood-chris:
Hi @zkamvar!

Thanks for the explanation :) I'm curious now where the `code-handout.R` file actually lives (if it's now not generated by `make_code_handout.R`), I can't see it in the repo!

zkamvar:
You don't see it because it's [generated when the lesson is built](https://carpentries.github.io/sandpaper/reference/build_handout.html), so it exists and is continually updated in the 
gh-pages branch (https://github.com/datacarpentry/R-ecology-lesson/blob/gh-pages/files/code-handout.R). 

See https://carpentries.github.io/sandpaper-docs/deployment.html for more details.

PR #882, opened 2023-12-05
Improve keyboard button visibility in before-we-start

ragamouf:
Currently, the <kbd> tags on the Ctrl key, Enter key, Cmd key all look black, or extremely dark grey. The text is also black. I don't think this is on purpose, and looking in the code there are backticks "`" which enclose the keyboard button text. Removing these backticks will hopefully allow the text to appear white, and visible on the screen.

The screenshot below shows what I see on my google chrome browser on a mac. 
Screenshot 2023-11-22 at 12 37 01 pm

I checked the code below and noticed that there were no backticks around the "F1" button.
Screenshot 2023-12-05 at 4 53 37 pm

### DIFF
--- a/episodes/00-before-we-start.Rmd
+++ b/episodes/00-before-we-start.Rmd
@@ -280,24 +280,24 @@ is a complete record of what we did, and anyone (including our future selves!)
 can easily replicate the results on their computer.
 
 RStudio allows you to execute commands directly from the script editor by using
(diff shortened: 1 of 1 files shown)


tobyhodges (approved):
Thank you for this @ragamouf. I am going to merge as this makes an important improvement to the readability of the lesson.

PR #919, opened 2025-04-15
Correct pivot GIF label where it should read long instead of wide

unode:
This seems like a typo as the wide label appears in all stages of the animation.

The PR goes from:

Old GIF pivot animation

to

New GIF pivot animation

where after the first animation the label was changed from **wide** to **long**.


Bisaloo (commented):
Nice catch and thanks for proposing a fix!

However, as far as I can tell, this file is no longer used in the lesson and the simplest fix might be to remove it. This simplifies the future maintenance load.

Were you using this file? How did you discover this issue? By browsing the source code, or is it used somewhere?

unode:
Yes, I was reviewing an older version of the lesson and looking for illustrations of pivoting. This one was useful.

unode:
Will you transfer the corrected animation there?

PR #926, opened 2025-05-16
Remove unused or outdated files

Bisaloo:
Most of the files became obsolete during the switch to the workbench, or during the lesson redesign.

One point I want to bring your attention on: https://github.com/datacarpentry/R-ecology-lesson/commit/4cbbf17d94d6aae53f499426e68d5d903e289939 overlaps with #919. I do think removing these files is the best solution if they're not used because we don't want to repo to accumulate fluff, but I may be missing something.


CarolynKoehn (changes requested):
I agree that it's a good idea to get rid of unused files!

CONDUCT.Rmd should not be removed -- it should be renamed to CODE_OF_CONDUCT.md. Otherwise the link to the code of conduct at the bottom of the lesson is broken. I checked the [spreadsheet-ecology-lesson](https://swcarpentry.github.io/r-novice-gapminder/CODE_OF_CONDUCT.html) and that's how it seems to work there.

Bisaloo:
I was still a bit reluctant to merge this as https://github.com/datacarpentry/R-ecology-lesson/pull/919 shows that community members are still using these files.
So I brought it up at the maintainers' meeting today and they encouraged us to indeed remove the files. The recommendation however is to try and get them into https://github.com/carpentries/commons before we pull the plug here.

Bisaloo:
> CONDUCT.Rmd should not be removed -- it should be renamed to CODE_OF_CONDUCT.md. Otherwise the link to the code of conduct at the bottom of the lesson is broken. I checked the [spreadsheet-ecology-lesson](https://swcarpentry.github.io/r-novice-gapminder/CODE_OF_CONDUCT.html) and that's how it seems to work there.

Thanks for the catch! I have moved this change to it own PR #931

CarolynKoehn:
Looks good to me!

PR #932, opened 2025-08-19
Fix Linux install instructions

### LINKED ISSUE #920: Install instructions for Linux are incorrect
### What is the problem?

They are a copy/paste of the instructions for macOS, but should be different:

https://github.com/datacarpentry/R-ecology-lesson/blob/27402dcccd10a11aa5d078f20ea968895de6a174/learners/setup.md?plain=1#L57-L66

### Location of problem (optional)

setup.md

Bisaloo:
Fix #920.

It is difficult to come up with something that precise for the general case as linux distributions come in many different flavours but I believe this PR is already a net improvement compared to the current situation, which gives invalid instructions.

### DIFF
--- a/learners/setup.md
+++ b/learners/setup.md
@@ -56,7 +56,7 @@ You have to install R before you install RStudio.
 
 ## For Linux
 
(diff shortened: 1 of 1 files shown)


CarolynKoehn (approved):
Looks good to me!

PR #943, opened 2025-10-07
Set seed at the start of the ggplot2 episode

Bisaloo:
This ensures we don't get unnecessary diffs when doing the monthly package update.

It is fine to have this in an included = FALSE chunk as we don't mind learners getting a different output. The only goal is facilitating infrastructure maintenance.


CarolynKoehn (approved):
Good idea!

PR #980, opened 2026-09-18
change outlier.shape = NA for outliers = FALSE

### LINKED ISSUE #979: Changing the way we teach how to remove outliers
### How could the content be improved?

I was just helping in a workshop during the ggplot episode, and reading the [geom_boxplot documentation](https://ggplot2.tidyverse.org/reference/geom_boxplot.html) I see there is a a relatively new (starting from 3.5 released in 2024) way of hiding outliers `outliers = FALSE`.

Probably that is a better way of teaching it than using the current `outlier.shape = NA`

### Which part of the content does your suggestion apply to?

https://datacarpentry.github.io/R-ecology-lesson/visualizing-ggplot.html#adding-geoms

josenino95:
_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._
Closes #979

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._
Explained in #979. But basically, in the ggplot episode, I changed all references from `outlier.shape = NA` to `outliers = FALSE`. I briefly read the text around the changes I made and I think I didn't brake anything in the sense of non-sensical text or references to the previous use of `outlier.shape = NA`

_If any relevant discussions have taken place elsewhere, please provide links to these._
See the issue linked above


episodes/visualizing-ggplot.Rmd
```diff
@@ -319,19 +319,19 @@ ggplot(data = complete_old, mapping = aes(x = plot_type, y = hindfoot_length)) +
   geom_jitter(alpha = 0.2)
 ```
 
-You may have noticed that some of our data points are now appearing on our plot twice: the outliers are plotted as black points from `geom_boxplot()`, but they are also plotted with `geom_jitter()`. Since we don't want to represent these data multiple times in the same form (points), we can stop `geom_boxplot()` from plotting them. We do this by setting the `outlier.shape` argument to `NA`, which means the outliers don't have a shape to be plotted.
+You may have noticed that some of our data points are now appearing on our plot twice: the outliers are plotted as black points from `geom_boxplot()`, but they are also plotted with `geom_jitter()`. Since we don't want to represent these data multiple times in the same form (points), we can stop `geom_boxplot()` from plotting them. We do this by setting the `outliers` argument to `FALSE`, which means the outliers don't have a shape to be plotted.
```
Bisaloo, line 322:
```suggestion
You may have noticed that some of our data points are now appearing on our plot twice: the outliers are plotted as black points from `geom_boxplot()`, but they are also plotted with `geom_jitter()`. Since we don't want to represent these data multiple times in the same form (points), we can stop `geom_boxplot()` from plotting them. We do this by setting the `outliers` argument to `FALSE`.
```

Bisaloo (changes requested):

Bisaloo (approved):
LGTM.

@datacarpentry/r-ecology-lesson-maintainers, could we have a second review if you're available please?

If not, I'll go ahead and merge this in the coming weeks.

